PR #14, opened 2018-06-21
DEVX-116: Java, Python, .NET, Go

ybyzek:


ccloud/ccloud-generate-cp-configs.sh
```diff
@@ -98,3 +133,259 @@ while read -r line
...
+using Confluent.Kafka;
+
+var producerConfig = new Dictionary<string, object>
+{
+    { "bootstrap.servers", "$BOOTSTRAP_SERVERS" },
+    { "broker.version.fallback", "0.10.0.0" },
+    { "api.version.fallback.ms", 0 },
+    { "sasl.mechanisms", "PLAIN" },
+    { "security.protocol", "SASL_SSL" },
+    { "ssl.ca.location", "/usr/local/etc/openssl/cert.pem" }, // linux, osx
```
edenhill: I don't believe this is a standard install path on Linux.
Also, why is this needed for the .NET client but not Python and Go? They should behave identically in this regard

ccloud/ccloud-generate-cp-configs.sh
```diff
@@ -98,3 +133,259 @@ while read -r line
...
+};
+
+var consumerConfig = new Dictionary<string, object>
+{
+    { "bootstrap.servers", "$BOOTSTRAP_SERVERS" },
+    { "broker.version.fallback", "0.10.0.0" },
+    { "api.version.fallback.ms", 0 },
+    { "sasl.mechanisms", "PLAIN" },
+    { "security.protocol", "SASL_SSL" },
+    { "ssl.ca.location", "/usr/local/etc/openssl/cert.pem" }, // linux, osx
```
edenhill: dito

ccloud/ccloud-generate-cp-configs.sh
```diff
@@ -98,3 +133,259 @@ while read -r line
...
+    topics: '<topic name>',
+    waitInterval: 0,
+    objectMode: false
+});
+EOF
+
+# C++
+CPP_CONFIG=$DEST/cpp.delta
+echo "$CPP_CONFIG"
+
```
edenhill: cppkafka is a community-maintained wrapper for librdkafka, is this really what we want to use?
We really don't know much about its quality, stability or governance.

I'd suggest using the librdkafka API directly, which is more verbose but at least it is something we support.

ccloud/ccloud-generate-cp-configs.sh
ybyzek: > I don't believe this is a standard install path on Linux.

This is verbatim taken from https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/examples/ConfluentCloud/Program.cs#L57 . I actually don't think we need to iterate on all OS types for the example, users should change the path for their system.  So I'm ok with no change here.

> Also, why is this needed for the .NET client but not Python and Go? They should behave identically in this regard

I don't know if {{ssl.ca.location}} omission is accidental or intentional from Python and Go examples for CCloud.

* Python: https://github.com/confluentinc/confluent-kafka-python/blob/master/examples/confluent_cloud.py
* Go: https://github.com/confluentinc/confluent-kafka-go/blob/master/examples/confluent_cloud_example/confluent_cloud_example.go

cc: @mhowlett for additional insight

ccloud/ccloud-generate-cp-configs.sh
ybyzek: @edenhill , this was unintentional on my part, total newbie mistake.  I will update the PR accordingly.  Thank you for the feedback.

FYI how I made the mistake: I saw `cppkafka` under https://github.com/edenhill/librdkafka/#language-bindings and went directly there.  I didn't see the librdkafka API examples for cpp until now.

ybyzek: @edenhill : I have updated the code to use `librdkafka` instead of `cppkafka`.  All feedback welcome.

PR #22, opened 2018-07-12
Migrate Oracle and Postgres demos from demo-scene

rmoff:
@ybyzek As discussed, I've created a `start.sh` for each, and brought the contents across as-is. Let me know if this works for you :)


ybyzek:
@rmoff : instead of `master`, want to target `5.0.x` branch?

PR #63, opened 2018-10-29
Update docker-compose.yml

agoujet:
Adding Control Center

### DIFF
--- a/ksql-workshop/docker-compose.yml
+++ b/ksql-workshop/docker-compose.yml
@@ -102,6 +102,31 @@ services:
                           bootstrap-server=kafka:29092 \
                           schemaRegistryUrl=http://schema-registry:8081'"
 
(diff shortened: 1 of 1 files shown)


rmoff:
Sorry I missed this PR. 
I'd prefer that we had separate docker-compose for C3. It's great including it but it adds resource requirements for people running the workshop, so maybe we make it optional by including it in a separate docker-compose. 
WDYT? Unless there's a clever way in docker-compose's syntax to specify a service but not start it at `up -d` and launch it separately? thinking out loud here…

ybyzek:
@rmoff can we still include C3 and absorb some of the performance optimizations in cp-demo?

ybyzek:
`ksql-workshop` example is no longer in the repo.  Closing PR

PR #73, opened 2018-12-05
DEVX-543: Confluent Python client examples to Confluent Cloud

ybyzek:


clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,94 @@
...
+    if len(sys.argv) != 3:
+        sys.stderr.write('Usage: %s <path to Confluent Cloud configuration file> <topic name>\n' % sys.argv[0])
+        sys.exit(1)
+
+    # Read Confluent Cloud configuration for librdkafka clients
+    config_file = sys.argv[1]
+    conf = {}
+    with open(config_file) as fh:
+      for line in fh:
+        if line[0] != "#" and line.strip():
```
edenhill: probably want to trim line before checks, whitespaces, like nettle, get in everywhere

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,94 @@
...
+          #print line
+          parameter, value = line.strip().split('=', 1)
+          conf[parameter] = value.strip()
+    fh.close
+    #print conf['bootstrap.servers']
```
edenhill: remove commented out line

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,94 @@
...
+    # Read Confluent Cloud configuration for librdkafka clients
+    config_file = sys.argv[1]
+    conf = {}
+    with open(config_file) as fh:
+      for line in fh:
+        if line[0] != "#" and line.strip():
+          #print line
+          parameter, value = line.strip().split('=', 1)
+          conf[parameter] = value.strip()
+    fh.close
```
edenhill: this does nothing, and is not needed since fh is in a context

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,94 @@
...
+    })
+
+    # Subscribe to topic
+    topic = sys.argv[2]
+    c.subscribe([topic])
+
+    total_count=0
+    try:
+        while True:
+            msg = c.poll(0.1)  # Wait for message or event/error
```
edenhill: 1.0 is a less CPU intensive timeout

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,94 @@
...
+            if msg is None:
+                # No message available within timeout.
+                # Initial message consumption may take up to `session.timeout.ms` for
+                #   the group to rebalance and start consuming
+                continue
+            elif not msg.error():
+                record_key = msg.key()
```
edenhill, line 68: Add a comment here saying this is a proper message (not an error), or switch the else to check if msg is error.

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,94 @@
...
+                record_value = msg.value()
+                print ("{0} \t {1}".format(record_key, record_value))
```
edenhill: the location, {0} and {1}, are not really necessary. {} will be fine.

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,94 @@
...
+                data = json.loads(record_value)
+                count = data['count']
+                total_count += count
+                print ("Updated total count to {0}".format(total_count))
```
edenhill: dito

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,94 @@
...
+
+            elif msg.error().code() == KafkaError._PARTITION_EOF:
```
edenhill: With the next version of confluent-kafka-python (and librdkafka), v1.0.0, partition eofs are no longer emitted, which simplifies the error checking

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,94 @@
...
+                print('end of partition: {0} [{1}] @ {2}'.format(msg.topic(), msg.partition(), msg.offset()))
+            else:
+                print('error: {0}'.format(msg.error().str()))
```
edenhill: no need for .str(), that's implicit.
drop {0} in favour of {}

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,99 @@
...
+    if len(sys.argv) != 3:
+        sys.stderr.write('Usage: %s <path to Confluent Cloud configuration file> <topic name>\n' % sys.argv[0])
+        sys.exit(1)
+
+    # Read Confluent Cloud configuration for librdkafka clients
+    config_file = sys.argv[1]
+    conf = {}
+    with open(config_file) as fh:
+      for line in fh:
+        if line[0] != "#" and line.strip():
```
edenhill: same: trim line
Also skip empty lines (do the same in the consumer)

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,99 @@
...
+          #print line
+          parameter, value = line.strip().split('=', 1)
+          conf[parameter] = value.strip()
+    fh.close
+    #print conf['bootstrap.servers']
```
edenhill: remove

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,99 @@
...
+    # Create topic if needed
+    topic = sys.argv[2]
+    a = AdminClient({
+           'bootstrap.servers': conf['bootstrap.servers'],
+           'sasl.mechanisms': 'PLAIN',
+           'security.protocol': 'SASL_SSL',
+           'sasl.username': conf['sasl.username'],
+           'sasl.password': conf['sasl.password']
+    })
+    fs = a.create_topics([NewTopic(topic, num_partitions=1, replication_factor=3)])
```
edenhill: wohoo, using the admin api! 💘

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,99 @@
...
+    for topic, f in fs.items():
+        try:
+            f.result()  # The result itself is None
+            print("Topic {} created".format(topic))
+        except Exception as e:
+            if e.args[0].code() != KafkaError.TOPIC_ALREADY_EXISTS:
+              print("Failed to create topic {}: {}".format(topic, e))
```
edenhill: add something to this printout to tell the user we accept this outcome.

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,99 @@
...
+        #    print("produced to: {0} [{1}] @ {2}".format(msg.topic(), msg.partition(), msg.offset()))
+
+    for n in range(10):
+      record_key = "alice"
+      data = {}
+      data['count'] = n
+      record_value = json.dumps(data)
+      print ("{0} \t {1}".format(record_key, record_value))
+      p.produce(topic,key=record_key,value=record_value,callback=acked)
+      p.poll(0)
```
edenhill: Add a comment above poll() saying something like "Serve delivery reports from previous produce() calls ..."

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,99 @@
...
+    def acked(err, msg):
+        """Delivery report callback called (from flush()) on successful or failed delivery of the message."""
+        if err is not None:
+            print("failed to deliver message: {0}".format(err.str()))
+        #else:
+        #    print("produced to: {0} [{1}] @ {2}".format(msg.topic(), msg.partition(), msg.offset()))
+
+    for n in range(10):
+      record_key = "alice"
+      data = {}
```
edenhill: you can instantiate the dict in one single line with the value: `data = {'count': n}`

edenhill (changes requested):

clients/cloud/python/consumer.py
ybyzek: Ack.  But sounds like no change for now

clients/cloud/python/producer.py
ybyzek: Woohoo!

clients/cloud/python/producer.py
ybyzek: @edenhill By delivery reports, do you mean callbacks?

clients/cloud/python/producer.py
ybyzek: I love shortcuts, thank you

gwenshap:
General nit: The location was non-obvious to me... maybe languages instead of clients? to indicate that we explore different languages here? And is there anything cloud specific about the example? (i.e. would we ever have a sub-directory that is called not-cloud?)

clients/cloud/python/README.md
```diff
@@ -0,0 +1,55 @@
+# Overview
+
+Produce messages to and consume messages from Confluent Cloud using Confluent Python Client for Apache Kafka
+
+
+# Prerequisites
+
+* Install Confluent's Python Client for Apache Kafka: https://github.com/confluentinc/confluent-kafka-python
+* File with client configuration parameters to connect to your Confluent Cloud instance
```
gwenshap: Link to instructions on how to get those? Especially the thing where there are tabs for different languages and that you need the C (librdkafka) one even though this is a python example.

clients/cloud/python/README.md
```diff
@@ -0,0 +1,55 @@
+# Overview
+
+Produce messages to and consume messages from Confluent Cloud using Confluent Python Client for Apache Kafka
+
+
+# Prerequisites
+
+* Install Confluent's Python Client for Apache Kafka: https://github.com/confluentinc/confluent-kafka-python
```
gwenshap: Does this work for any version of the client? If not, we need to specify a version (old versions of librdkafka had different configurations, IIRC).

clients/cloud/python/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+# Usage
+
+$ cat ~/.ccloud/librdkafka.config
+bootstrap.servers=<>
+sasl.username=<>
+sasl.password=<>
+
+
+$ ./producer.py ~/.ccloud/librdkafka.config test1
```
gwenshap: I got the following output:
```

%3|1544080882.161|FAIL|rdkafka#producer-1| [thrd:sasl_ssl://b1-pkc-epvye.us-central1.gcp.confluent.cloud:9092/1]: sasl_ssl://b1-pkc-epvye.us-central1.gcp.confluent.cloud:9092/1: SASL authentication failure: Disconnected: check client PLAIN credentials and broker logs
%3|1544080882.161|ERROR|rdkafka#producer-1| [thrd:sasl_ssl://b1-pkc-epvye.us-central1.gcp.confluent.cloud:9092/1]: sasl_ssl://b1-pkc-epvye.us-central1.gcp.confluent.cloud:9092/1: SASL authentication failure: Disconnected: check client PLAIN credentials and broker logs
%3|1544080882.164|FAIL|rdkafka#producer-2| [thrd:sasl_ssl://b1-pkc-epvye.us-central1.gcp.confluent.cloud:9092/1]: sasl_ssl://b1-pkc-epvye.us-central1.gcp.confluent.cloud:9092/1: SASL authentication failure: Disconnected: check client PLAIN credentials and broker logs
%3|1544080882.164|ERROR|rdkafka#producer-2| [thrd:sasl_ssl://b1-pkc-epvye.us-central1.gcp.confluent.cloud:9092/1]: sasl_ssl://b1-pkc-epvye.us-central1.gcp.confluent.cloud:9092/1: SASL authentication failure: Disconnected: check client PLAIN credentials and broker logs
Topic test1 created
alice 	 {"count": 0}
alice 	 {"count": 1}
alice 	 {"count": 2}
alice 	 {"count": 3}
alice 	 {"count": 4}
alice 	 {"count": 5}
alice 	 {"count": 6}
alice 	 {"count": 7}
alice 	 {"count": 8}
alice 	 {"count": 9}
%3|1544080884.647|FAIL|rdkafka#producer-1| [thrd:sasl_ssl://b1-pkc-epvye.us-central1.gcp.confluent.cloud:9092/1]: sasl_ssl://b1-pkc-epvye.us-central1.gcp.confluent.cloud:9092/1: SASL authentication failure: Disconnected: check client PLAIN credentials and broker logs
%3|1544080884.647|ERROR|rdkafka#producer-1| [thrd:sasl_ssl://b1-pkc-epvye.us-central1.gcp.confluent.cloud:9092/1]: sasl_ssl://b1-pkc-epvye.us-central1.gcp.confluent.cloud:9092/1: SASL authentication failure: Disconnected: check client PLAIN credentials and broker logs
```

clients/cloud/python/README.md
```diff
@@ -0,0 +1,55 @@
...
+alice 	 {"count": 3}
+alice 	 {"count": 4}
+alice 	 {"count": 5}
+alice 	 {"count": 6}
+alice 	 {"count": 7}
+alice 	 {"count": 8}
+alice 	 {"count": 9}
+
+
+$ ./consumer.py ~/.ccloud/librdkafka.config test1
```
gwenshap: And... 
```
gwenshap-mac:python gwen$ ./consumer.py ~/.ccloud/librdkafka.config test1
end of partition: test1 [0] @ 10
```
It looks like I produced 10 events (otherwise, why offset 10), but somehow didn't consume them?

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,91 @@
...
+      for line in fh:
+        line = line.strip()
+        if line[0] != "#" and len(line) != 0:
+          parameter, value = line.strip().split('=', 1)
+          conf[parameter] = value.strip()
+
+    # Create Consumer instance
+    c = Consumer({
+        'bootstrap.servers': conf['bootstrap.servers'],
+        'broker.version.fallback': '0.10.0.0',
```
gwenshap: What does this mean?

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,91 @@
...
+        'api.version.fallback.ms': 0,
+        'sasl.mechanisms': 'PLAIN',
+        'security.protocol': 'SASL_SSL',
+        'sasl.username': conf['sasl.username'],
+        'sasl.password': conf['sasl.password'],
+        'group.id': str(uuid.uuid1()),  # this will create a new consumer group on each invocation.
```
gwenshap: A real application would basically never do that, right? so I'm wondering why do we do this in an example. If we want to give the users an easy way to start from scratch, wouldn't a command-line parameter be a better option?

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # Subscribe to topic
+    topic = sys.argv[2]
+    c.subscribe([topic])
+
+    total_count=0
+    try:
+        while True:
+            msg = c.poll(1.0)  # Wait for message or event/error
+            if msg is None:
```
gwenshap, line 60: Do we want to print something so users will see what is going on? "waiting for events..."

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,91 @@
...
+                continue
+            elif not msg.error():
+                # Check for Kafka message
+                record_key = msg.key()
+                record_value = msg.value()
+                print ("{} \t {}".format(record_key, record_value))
+                data = json.loads(record_value)
+                count = data['count']
+                total_count += count
+                print ("Updated total count to {}".format(total_count))
```
gwenshap: Do you want to commit explicitly at some point? Or mention the existence of commits and why you can safely ignore them here?

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,96 @@
...
+# =============================================================================
+
+from confluent_kafka import Producer, KafkaError
+from confluent_kafka.admin import AdminClient, NewTopic
+import sys
+import json
+
+
+if __name__ == '__main__':
+    if len(sys.argv) != 3:
```
gwenshap: Maybe worth checking if Python has a nicer command argument parser (@rnpridgeon will know). Mostly because it will make the example code a bit more similar to what one should run in production.

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,96 @@
...
+
+    # Read Confluent Cloud configuration for librdkafka clients
+    config_file = sys.argv[1]
+    conf = {}
+    with open(config_file) as fh:
+      for line in fh:
+        line = line.strip()
+        if line[0] != "#" and len(line) != 0:
+          parameter, value = line.strip().split('=', 1)
+          conf[parameter] = value.strip()
```
gwenshap: I wonder if this is worth printing out (in the consumer too).

Or if librdkafka prints it in a different debug level, maybe document how to set the debug level in Python.

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,96 @@
...
+           'sasl.password': conf['sasl.password']
+    })
+    fs = a.create_topics([NewTopic(topic, num_partitions=1, replication_factor=3)])
+    for topic, f in fs.items():
+        try:
+            f.result()  # The result itself is None
+            print("Topic {} created".format(topic))
+        except Exception as e:
+            # Continue if error code TOPIC_ALREADY_EXISTS, which may be valid if topic exists
+            if e.args[0].code() != KafkaError.TOPIC_ALREADY_EXISTS:
```
gwenshap, line 69: Nice bit of real-world error-handling.

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,96 @@
...
+              print("Failed to create topic {}: {}".format(topic, e))
+
+    # Optional per-message delivery callback (triggered by poll() or flush())
+    # when a message has been successfully delivered or permanently
+    # failed delivery (after retries).
+    def acked(err, msg):
+        """Delivery report callback called (from flush()) on successful or failed delivery of the message."""
+        if err is not None:
+            print("failed to deliver message: {0}".format(err.str()))
+        #else:
```
gwenshap: Why is this commented out?

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,96 @@
...
+        #    print("produced to: {0} [{1}] @ {2}".format(msg.topic(), msg.partition(), msg.offset()))
+
+    for n in range(10):
+      record_key = "alice"
+      record_value = json.dumps({'count': n})
+      print ("{0} \t {1}".format(record_key, record_value))
```
gwenshap: Indicate that you are printing before producing. I took the output as indication that we produced successfully.

clients/cloud/python/producer.py
gwenshap: If you want to mimic real world handling of success, up a counter and after flushing print "N messages were successfully produced. Good job! Now go consume them".

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,96 @@
...
+      p.produce(topic,key=record_key,value=record_value,callback=acked)
```
gwenshap: I'm pretty sure that the callback does not work as intended (I got errors, I didn't produce messages, and I never saw the callback printout).

clients/cloud/python/README.md
```diff
@@ -0,0 +1,55 @@
+# Overview
+
+Produce messages to and consume messages from Confluent Cloud using Confluent Python Client for Apache Kafka
```
gwenshap: Since my inspiration for the project is this: https://www.rabbitmq.com/tutorials/tutorial-one-python.html
I'd love to see an explanation that places this example in the grand scheme of things and leaves place for other examples (many consumers for instance).

gwenshap (changes requested):
I'll still debug why this doesn't work for me (Admin client seems to work, but producer didn't? Or maybe consumer didn't?).

gwenshap:
Running it again and producer had no errors at all, so it may be related to the admin client?

gwenshap:
Ah, only took me an hour to figure this out. I guess I have a different version of our Python client than you (although I ran "pip install --upgrade pip && pip install confluent-kafka" in an attempt to bump it.

Anyway, my consumer was consuming from latest, so it didn't see any of the messages we produced. I had to add the following configuration to make reset work:

```
'default.topic.config': {'auto.offset.reset': 'smallest'},
```

On the way, I learned how to enable debug logs. You may want to add this to the example? 
```
'debug': 'consumer'
```

(btw, did you spot the intense foreshadowing from an early comment in the review, mentioning that old versions have different configuration? and I still fell for it!)

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,91 @@
...
+
+            elif msg.error().code() == KafkaError._PARTITION_EOF:
+                print('end of partition: {0} [{1}] @ {2}'.format(msg.topic(), msg.partition(), msg.offset()))
```
gwenshap: I think there is no need to specify the positions. {} should work?

clients/cloud/java/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+1. Now we can turn our attention to the client examples in this directory.
+
+	1. First run the example producer to publish 10 data records to Confluent Cloud.
+
+		```shell
+		# Build the client examples
+		$ mvn clean package
+		
+		# Run the producer
```
rnpridgeon, line 56: Why not provide a runner script like `kafka-console-producer.sh` instead of making folks type out a long winded mvn command?

rnpridgeon: This entire `README`  has nothing to do with the python client either, should it be a part of another PR?

clients/cloud/python/README.md
rnpridgeon: seems like an excellent chance to demonstrate the use of Python's configurable logger!

Python stdlib logger examples:
https://docs.python.org/2/howto/logging.html#logging-to-a-file

` producer = confluent_kafka.Producer({‘bootstrap.servers’: ‘mybroker:29092’}, logger=mylogger)`

clients/cloud/python/README.md
rnpridgeon: `auto.offset.reset` defaults to latest. We could use `assign` or just default to `earliest`. Alternatively the demo could start the consumer first, place it in the background. 

`./consumer.py ~/.ccloud/librdkafka.config test1 &`

clients/cloud/python/consumer.py
rnpridgeon: @gwenshap this is how older librdkafka clients negotiated versions with the broker (before you could query the broker for the version) 

@ybyzek I would remove  `broker.version.fallback` and `api.version.fallback.ms=0`. This combination is actually a workaround for a race condition which can cause librdkafka to think the broker is of an older version than it really is. As a result headers could be dropped silently causing much sadness. Either way it's not something people should just be copying and pasting and its unnecessary if you aren't producing with headers.

clients/cloud/python/consumer.py
rnpridgeon: +1, we do this in integration tests but it should not be used in the real world. Examples should be something you can essentially copy and paste.

clients/cloud/python/consumer.py
rnpridgeon: 1.0 will be out soon and all of our examples/tests should reflect the new behavior to avoid setting the wrong expectation. This will undoubtedly cause confusion and likely support cases. I recommend either removing it or at the very least explicitly adding `enable.partition.eof=true` to the consumer configuration.

clients/cloud/python/producer.py
rnpridgeon: `argparse` isn't amazing but it is part of the standard library and its better than parsing the arguments yourself. 

https://github.com/confluentinc/confluent-kafka-python/blob/master/examples/avro-cli.py#L172-L189

clients/cloud/python/producer.py
rnpridgeon: Use `on_delivery` instead. They do the same exact thing but `on_delivery` is more descriptive and ideally I'd like to deprecate and remove `callback` in the future. Until then I want to heavily encourage people to use `on_delivery`.

clients/cloud/python/README.md
ybyzek: Ack

clients/cloud/python/producer.py
rnpridgeon: `Delivery reports` are what your callback is handling. When you send a produce request an additional object is created, `msgstate`. As the name implies this stores the state of each message you produced.  Once the produce request completes the message state is updated and placed on the consumer queue. Each time you call poll() these reports will be served so whatever logic you put in the handler can be executed. Most examples seem to use this for informative purposes but you could also do some sort of cool pipelining with it.

clients/cloud/python/README.md
ybyzek: @edenhill 

1. Can you please advise if there are version caveats/constraints that we need to call out?  
2. If there are caveats, in general it may make sense to capture those NOT in this README but rather in the README in https://github.com/confluentinc/confluent-kafka-python , which should be the source of truth

clients/cloud/python/consumer.py
ybyzek: From https://docs.confluent.io/current/cloud/using/config-client.html#librdkafka-based-c-clients

> The api.version.request, broker.version.fallback, and api.version.fallback.ms options instruct librdkafka to use the latest protocol version and not fallback to an older version.

clients/cloud/python/producer.py
ybyzek: @gwenshap can you please elaborate, when you write "this is worth printing out" -- what is _this_?

In case you mean the `bootstrap.server`, `sasl.username`, and `sasl.password`, I actually deliberately didn't print that out to keep api key/password private and not share in stdout

If you mean something else, please let me know.

clients/cloud/python/producer.py
ybyzek: Made possible by the great work of the clients team to add AdminClient functionality into the client!  Thanks @edenhill !

clients/cloud/python/consumer.py
ybyzek: Ack

clients/cloud/java/README.md
ybyzek: I moved the java example to a `java` subdirectory, so we can ignore that code for now.  It's already part of the repo.

clients/cloud/python/producer.py
ybyzek: Ack

clients/cloud/python/producer.py
ybyzek: Thanks for the clarification.  Comment added

clients/cloud/python/consumer.py
edenhill: >  I would remove broker.version.fallback and api.version.fallback.ms=0. 

Don't remove them, they're needed to make sure the latest protocol version is used.
Otherwise the broker will need to perform up/down conversion which slows things down.

clients/cloud/python/producer.py
ybyzek: Ack

ybyzek:
@gwenshap , @rnpridgeon the consumer code already had set 

```        'auto.offset.reset': 'earliest'```

https://github.com/confluentinc/examples/blob/DEVX-543/clients/cloud/python/consumer.py#L56

This is at odds with the behavior @gwenshap encountered.

File: clients/cloud/python/README.md (hunk 18 above)
Comment (ybyzek) [reply clients/cloud/python/README.md]: It feels like logging should be out of scope.  This example is meant to be simple and narrowly focused on client-Cloud interaction, not showcase of Python best practices.

File: clients/cloud/python/README.md (hunk 30 above)
Comment (ybyzek) [reply clients/cloud/python/README.md]: I added a bit more into the README.  I figure we can expand on the explanations later as we flesh out the ideas with the blog posts.

File: clients/cloud/python/README.md (hunk 19 above)
Comment (ybyzek) [reply clients/cloud/python/README.md]: Sorry for dupe comment, but adding here in the thread (I couldn't comment here before...it was strange)

The consumer code already had set 

```bash
'auto.offset.reset': 'earliest'
```

https://github.com/confluentinc/examples/blob/DEVX-543/clients/cloud/python/consumer.py#L56

This is at odds with the behavior @gwenshap encountered, because the consumer should have started at the beginning of the log.

@edenhill , @rnpridgeon : is there some difference in behavior between versions that would explain this?

ybyzek:
@gwenshap FWIW, this is how I installed

```bash
pip install --no-cache-dir --no-binary :all: confluent-kafka
```

ybyzek:
@gwenshap FWIW, this is how I installed

```bash
pip install --no-cache-dir --no-binary :all: confluent-kafka```

**Comment by ybyzek** [comment]
@gwenshap FWIW, this is how I installed

```bash
pip install --no-cache-dir --no-binary :all: confluent-kafka```

**Comment by ybyzek** [comment]
@gwenshap FWIW, this is how I installed

```bash
pip install --no-cache-dir --no-binary :all: confluent-kafka```

File: clients/cloud/python/producer.py (hunk 29 above)
Comment (ybyzek) [reply clients/cloud/python/producer.py]: Ack and done!

File: clients/cloud/python/consumer.py (hunk 21 above)
Comment (ybyzek) [reply clients/cloud/python/consumer.py]: @rnpridgeon I agree with "Examples should be something you can essentially copy and paste."

Evidence: this is straight copy/paste from the Confluent Kafka Python examples:

https://github.com/confluentinc/confluent-kafka-python/blob/34baea620f063cb96fc32b5a562170f19a19228e/examples/confluent_cloud.py#L89

I can remove it from this example, but someone needs to fix there

File: clients/cloud/python/consumer.py (hunk 22 above)
Comment (ybyzek) [reply clients/cloud/python/consumer.py:60]: Ack

File: clients/cloud/python/consumer.py (hunk 21 above)
Comment (rnpridgeon) [reply clients/cloud/python/consumer.py]: Good catch, will fix

**Comment by ybyzek** [comment]
@rnpridgeon FYI I switched to `argparse` and it *is* amazing!  Thanks for the suggestion

File: clients/cloud/python/README.md (hunk 18 above)
Comment (gwenshap) [reply clients/cloud/python/README.md]: I'm with @rnpridgeon: Where not overly complex, the example should show what good client code looks like. This one-line is something I'd want to know about (how the logger is passed to the producer).

If we are fair, the cloud part is only: Copy-paste stuff from the web and read configuration from a file. The rest is "how to use Confluent-python client", and getting the logs is an important part IMO.

File: clients/cloud/python/README.md (hunk 19 above)
Comment (gwenshap) [reply clients/cloud/python/README.md]: Note that the issue here was fixed by adding:        
```
 'default.topic.config': {'auto.offset.reset': 'smallest'}
```
to my configuration

File: clients/cloud/python/consumer.py (hunk 20 above)
Comment (gwenshap) [reply clients/cloud/python/consumer.py]: I don't get it. The version in the cloud is 1.1, right? And we are assuming a new client too (because admin client), so why are we specifying 0.10.0.0 as a version?

If this confuses me (and I have few bits of Kafka experience), at the very least we need to document why this is here.

File: clients/cloud/python/producer.py (hunk 25 above)
Comment (gwenshap) [reply clients/cloud/python/producer.py]: I meant printing out the config (without the security bits) - the consumer group part can be important for instance. 

No strong feelings, but Java does log configs.

File: clients/cloud/python/README.md (hunk 19 above)
Comment (gwenshap) [reply clients/cloud/python/README.md]: My issue occurred with confluent-kafka 0.11.5. It was resolved (without the extra config) when I upgraded to 0.11.6.

I suggest documenting that this example works with confluent-kafka 0.11.6.

File: clients/cloud/python/README.md [hunk 33]
```diff
@@ -0,0 +1,67 @@
...
+Preparing to produce record: alice 	 {"count": 1}
+Preparing to produce record: alice 	 {"count": 2}
+Preparing to produce record: alice 	 {"count": 3}
+Preparing to produce record: alice 	 {"count": 4}
+Preparing to produce record: alice 	 {"count": 5}
+Preparing to produce record: alice 	 {"count": 6}
+Preparing to produce record: alice 	 {"count": 7}
+Preparing to produce record: alice 	 {"count": 8}
+Preparing to produce record: alice 	 {"count": 9}
+Successfully produced record to topic test1 partition [0] @ offset None
```
edenhill: This printout doesn't look right, there should a proper offset printed.
What py client and librdkafka versions are you on?

clients/cloud/python/consumer.py
edenhill: By setting `broker.version.fallback` to a AK version that supports ApiVersionRequest we're forcing the client to always issue ApiVersionRequests, even if the heuristics (connection setup failure, timeout, etc) deems it should downgrade to an older version.
This is for seamlessly supporting AK 0.8 and 0.9 clusters, which I don't think is a sensible default anymore.

I think we'll take the chance to improve these semantics for the v1.0.0 release, in which case neither of these configs will be needed, so I suggest you remove the following configs: broker.version.fallback, api.version.fallback.ms.

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,95 @@
...
+            elif not msg.error():
+                # Check for Kafka message
+                record_key = msg.key()
+                record_value = msg.value()
+                data = json.loads(record_value)
+                count = data['count']
+                total_count += count
+                print ("Consumed record with key {} and value {}, and updated total count to {}".format(record_key,record_value,total_count))
+            else:
+                if msg.error().code() != KafkaError._PARTITION_EOF:
```
edenhill: enable.partition.eof is changed to false, thus no PARTITION_EOF "errors", in v1.0.0.

If we can tell people to use confluent-kafka-python v1.0.0 for these examples they can be simplified.

clients/cloud/python/producer.py
edenhill: My strong opinion is that the Java client shouldn't print anything unless explicitly told to, since it is a library, not an application.
In this case we're making an application, but I'd consider the default printing of config an anti-pattern, it doesn't serve much purpose and just gets in the way, possibly exposing secrets in (CI, et.al) output.

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,104 @@
...
+            print("failed to deliver message: {0}".format(err.str()))
+        else:
+            print("Successfully produced record to topic {} partition [{}] @ offset {}".format(msg.topic(), msg.partition(), msg.offset()))
+
+    for n in range(10):
+      record_key = "alice"
+      record_value = json.dumps({'count': n})
+      print ("Preparing to produce record: {} \t {}".format(record_key, record_value))
+      p.produce(topic,key=record_key,value=record_value,on_delivery=acked)
+      # p.poll(0) serves delivery reports from previous produce() calls to the on_delivery handler
```
edenhill: This doesn't read that well in my opinion

`# p.poll(0) serves delivery reports (on_delivery) from previous produce() calls.`

edenhill (changes requested):
as for logger integration: I too believe that is outside the scope of this example.

clients/cloud/python/consumer.py
ybyzek: Ack.  Removing EOF "error" and recommending v1.0

clients/cloud/python/consumer.py
ybyzek: Ack, removing `broker.version.fallback`, `api.version.fallback.ms`.  Yay for simplification!

clients/cloud/python/README.md
ybyzek: Now requiring v1.0 for this example

clients/cloud/python/producer.py
ybyzek: Ack

clients/cloud/python/README.md
ybyzek: Recommendation is going to be 1.0

clients/cloud/python/README.md
ybyzek: As discussed internally, we are omitting logging

clients/cloud/python/producer.py
ybyzek: Leaving this out

clients/cloud/python/README.md
ybyzek: Resolved in internal discussion

clients/cloud/python/ccloud_lib.py
```diff
@@ -0,0 +1,50 @@
...
+    required.add_argument('-f', dest="config_file", help="path to Confluent Cloud configuration file", required=True)
+    required.add_argument('-t', dest="topic", help="topic name", required=True)
+    args = parser.parse_args()
+ 
+    return args
+
+
+def read_ccloud_config(config_file):
+
+    # Read Confluent Cloud configuration for librdkafka clients
```
edenhill: This should be a docstring directly following the function line:
```
def read..():
    """ Read Confluent cloud ... """
```

clients/cloud/python/ccloud_lib.py
```diff
@@ -0,0 +1,50 @@
...
+# =============================================================================
+#
+# Helper module
+#
+# =============================================================================
+
+import argparse
+
+def parse_args():
+
```
edenhill, line 29: docstring: """ Parse command line arguments """

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,80 @@
...
+    conf = ccloud_lib.read_ccloud_config(config_file)
+
+    # Create Consumer instance
+    c = Consumer({
+        'bootstrap.servers': conf['bootstrap.servers'],
+        'sasl.mechanisms': 'PLAIN',
+        'security.protocol': 'SASL_SSL',
+        'sasl.username': conf['sasl.username'],
+        'sasl.password': conf['sasl.password'],
+        'group.id': 'group1',
```
edenhill: might be better to use 'example_group1' to avoid joining an exisitng production group, unlikely as it is, it also shows that this value can/should be changed.

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,80 @@
...
+        'auto.offset.reset': 'earliest'
```
edenhill, line 48: I like comments, like: # Start reading from the beginning of the topic if no committed offsets exist

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,80 @@
...
+    })
+
+    # Subscribe to topic
+    c.subscribe([topic])
+
+    # Process messages
+    total_count=0
+    try:
+        while True:
+            print "Waiting for message or event/error in poll()"
```
edenhill: print("..") for py3 compliance

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,80 @@
...
+                #   the group to rebalance and start consuming
+                continue
+            elif not msg.error():
+                # Check for Kafka message
+                record_key = msg.key()
+                record_value = msg.value()
+                data = json.loads(record_value)
+                count = data['count']
+                total_count += count
+                print ("Consumed record with key {} and value {}, and updated total count to {}".format(record_key,record_value,total_count))
```
edenhill: remove whitespace between print and (, for consistency

clients/cloud/python/consumer.py
```diff
@@ -0,0 +1,80 @@
...
+            else:
+                print('error: {}'.format(msg.error()))
```
edenhill, line 77: use double quotes, for consistency, and capitalize the sentence

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,88 @@
...
+#
+# =============================================================================
+
+from confluent_kafka import Producer, KafkaError
+from confluent_kafka.admin import AdminClient, NewTopic
+import json
+import ccloud_lib
+
+
+
```
edenhill: `pip install flake8`, then run `flake8` from the directory containing these files to make sure they are formatted correctly.

Also verify these scripts with python3 (`virtualenv -p python3.6 ~/venv3 ; source ~/venv3/bin/activate`)

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,88 @@
...
+    # Create Producer instance
+    p = Producer({
+           'bootstrap.servers': conf['bootstrap.servers'],
+           'sasl.mechanisms': 'PLAIN',
+           'security.protocol': 'SASL_SSL',
+           'sasl.username': conf['sasl.username'],
+           'sasl.password': conf['sasl.password']
+    })
+
+    # Create topic if needed
```
edenhill, line 48: I would move this to its own function, and create the topic before the Producer.

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,88 @@
...
+            if e.args[0].code() != KafkaError.TOPIC_ALREADY_EXISTS:
+              print("Failed to create topic {}: {}".format(topic, e))
+
+    # Optional per-message on_delivery handler (triggered by poll() or flush())
+    # when a message has been successfully delivered or
+    # permanently failed delivery (after retries).
+    def acked(err, msg):
+        """Delivery report handler called on successful or failed delivery of the message."""
+        if err is not None:
+            print("failed to deliver message: {0}".format(err.str()))
```
edenhill: capitalize on the failed

edenhill: drop position formatter: {0} -> {}

edenhill: You don't need to call str() on the error, from what I understand.

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,88 @@
...
+        else:
+            print("Successfully produced record to topic {} partition [{}] @ offset {}".format(msg.topic(), msg.partition(), msg.offset()))
+
+    for n in range(10):
+      record_key = "alice"
+      record_value = json.dumps({'count': n})
+      print ("Preparing to produce record: {} \t {}".format(record_key, record_value))
```
edenhill: remove whitespace

edenhill: "Producing" reads better, Preparing sounds like a multi-step thing

clients/cloud/python/producer.py
```diff
@@ -0,0 +1,88 @@
...
+      p.produce(topic,key=record_key,value=record_value,on_delivery=acked)
+      # p.poll(0) serves delivery reports (on_delivery) from previous produce() calls.
```
edenhill: remove 0, might give people the idea that it only does this if a timeout of 0 is provided

edenhill (changes requested):
Verify that it works with Python 3, and passes flake8

clients/cloud/python/producer.py
ybyzek: I disagree, I think AdminClient is a key feature of the python client and topic creation is required for Confluent Cloud.  So it's important to showcase prominently

clients/cloud/python/producer.py
ybyzek: Ack on all three

clients/cloud/python/producer.py
ybyzek: @edenhill Done: reformatted with flake8 and verified python3.7

edenhill (approved):
Great stuff!

ybyzek:
Thanks for all the great feedback!  Merging...

PR #76, opened 2018-12-08
DEVX-552: Add Ruby client examples to Confluent Cloud

stanislavkozlovski:
Followed the example from #73 

cc @ybyzek @mhowlett @gAmUssA @edenhill @gwenshap @rnpridgeon for review


clients/cloud/README.md
```diff
@@ -4,6 +4,7 @@ This directory includes examples of Kafka client streaming applications connecti
 
 [![](images/java.png)](java/)
 [![](images/python.png)](python/)
+[![](images/ruby.png)](ruby/)
```
ybyzek, line 7: Resize `images/ruby.png` to be a bit smaller?

ybyzek (approved):
Thanks for the PR @stanislavkozlovski !

LGTM from my perspective, but please don't merge until review/approval from the clients team.

clients/cloud/ruby/README.md
```diff
@@ -0,0 +1,68 @@
...
+* Local file with configuration parameters to connect to your Confluent Cloud instance ([how do I find those?](https://docs.confluent.io/current/cloud/using/config-client.html#librdkafka-based-c-clients)). Format the file as follows:
+    ```bash
+    $ cat ~/.ccloud/ccloud.config
+    bootstrap.servers=<broker-1,broker-2,broker-3>
+    sasl.username=<api-key-id>
+    sasl.password=<secret-access-key>
+    ```
+* Install gems
+    ```bash
+    $ bundle install
```
edenhill, line 20: What directly should this be run from?

clients/cloud/ruby/c_cloud.rb
```diff
@@ -0,0 +1,100 @@
...
+# Helper class for working with Confluent Cloud
+#
+class CCloud
+  attr_reader :kafka
+
+  def initialize
+    # Parse arguments, load the CCloud config and initialize the Kafka client
+    @args = parse_args!
+    @config = load_config(@args[:config])
+    @kafka = Kafka.new(
```
edenhill, line 37: What is telling the client to connect with SSL?

clients/cloud/ruby/c_cloud.rb
```diff
@@ -0,0 +1,100 @@
...
+    @args[:topic]
+  end
+
+  def consumer
+    return @consumer unless @consumer.nil?
+
+    # Consumers with the same group id will form a Consumer Group together.
+    @consumer = @kafka.consumer(group_id: 'ruby_example_group_1')
+
+    # Stop the consumer when the SIGTERM signal is sent to the process.
```
edenhill: Since this runs in the foreground I'd suggest SIGINT (ctrl-c) instead (or both)

clients/cloud/ruby/c_cloud.rb
```diff
@@ -0,0 +1,100 @@
...
+    @consumer
+  end
+
+  private
+
+  def parse_args!
+    options = {}
+
+    OptionParser.new do |opts|
+      opts.banner = 'Confluent Python Client example to produce messages to Confluent Cloud'
```
edenhill: Python -> Ruby

edenhill: Since this is used for the consumer it probably shouldn't say "produce messages" here

clients/cloud/ruby/consumer.rb
```diff
@@ -0,0 +1,52 @@
...
+    record_value = message.value
+    data = JSON.parse(record_value)
+    count = data['count']
+    total_count += count
+
+    puts "Consumed record with key #{record_key} and value #{record_value}, " \
+         "and updated total count #{total_count}"
+  end
+rescue Kafka::Error => e
+  puts "Consuming messages from #{topic} failed: #{e.message}"
```
edenhill: I don't know ruby, so does this exit the loop or continue going? (should be the latter)

clients/cloud/ruby/producer.rb
```diff
@@ -0,0 +1,50 @@
...
+
+require 'json'
+require './c_cloud'
+
+ccloud = CCloud.new
+topic = ccloud.topic
+begin
+  ccloud.kafka.create_topic(topic, num_partitions: 1, replication_factor: 3)
+  puts "Created topic #{topic}"
+rescue Kafka::Error => e
```
edenhill: This needs to handle TOPIC_ALREADY_EXISTS, right?

clients/cloud/ruby/producer.rb
```diff
@@ -0,0 +1,50 @@
...
+end
+
+0.upto(9).each do |n|
+  record_key = 'alice'
+  record_value = JSON.dump(count: n)
+  record = "#{record_key}\t#{record_value}"
+  puts "Producing record: #{record}"
+
+  begin
+    ccloud.kafka.deliver_message(record_value, key: record_key, topic: topic)
```
edenhill: deliver_message() is a synchronous send, which is fine for this example but generally not what people should be using since it will absolutely kill performance.

I'd like for these examples to "do the right thing" and use an asynchronous producer.

edenhill (changes requested):
Nice stuff, bunch of comments in line.

clients/cloud/ruby/README.md
stanislavkozlovski: From the `/ruby` directory. I'm unsure on how to best specify this

clients/cloud/ruby/c_cloud.rb
stanislavkozlovski: the `ssl_ca_certs_from_system` arg

From the zendesk docs:

> Authentication using SASL
Kafka has support for using SASL to authenticate clients. Currently GSSAPI, SCRAM and PLAIN mechanisms are supported by ruby-kafka.

> NOTE: With SASL for authentication, it is highly recommended to use SSL encryption. The default behavior of ruby-kafka enforces you to use SSL and you need to configure SSL encryption by passing ssl_ca_cert or enabling ssl_ca_certs_from_system. However, this strict SSL mode check can be disabled by setting sasl_over_ssl to false while initializing the client.

clients/cloud/ruby/c_cloud.rb
stanislavkozlovski: This actually catches all and is a left-over comment - so I'll just remove it

clients/cloud/ruby/c_cloud.rb
stanislavkozlovski: Oh wow, whoops

clients/cloud/ruby/consumer.rb
stanislavkozlovski: Ugh, yeah... it seems to re-raise every error. Should we just keep retrying infinitely?

clients/cloud/ruby/producer.rb
stanislavkozlovski: Yeah, we should fail on other errors 👍

clients/cloud/ruby/producer.rb
stanislavkozlovski: I was thinking about this and decided to go with the simpler approach for the example. I don't feel strongly about this - i'll switch to the async producer

clients/cloud/ruby/c_cloud.rb
```diff
@@ -0,0 +1,112 @@
...
+
+  def topic
+    @args[:topic]
+  end
+
+  #
+  # An asynchronous producer tweaked for the low-throughput needs of this example
+  #
+  def producer
+    @producer ||= @kafka.async_producer(
```
stanislavkozlovski, line 53: `||=` initializes `@producer` if it's nil, otherwise doesn't do anything (always returns `@producer`)

clients/cloud/ruby/README.md
```diff
@@ -0,0 +1,68 @@
...
+Produce messages to and consume messages from [Confluent Cloud](https://www.confluent.io/confluent-cloud/) using the [ ZenDesk Ruby Client for Apache Kafka](https://github.com/zendesk/ruby-kafka).
+
+
+# Prerequisites
+
+* [Bundler](https://bundler.io/) installed on your machine. Install via `gem install bundler`
+* Access to a [Confluent Cloud](https://www.confluent.io/confluent-cloud/) cluster
+* Local file with configuration parameters to connect to your Confluent Cloud instance ([how do I find those?](https://docs.confluent.io/current/cloud/using/config-client.html#librdkafka-based-c-clients)). Format the file as follows:
+    ```bash
+    $ cat ~/.ccloud/ccloud.config
```
edenhill: Should be librdkafka.config to avoid ambiguity with `~/.ccloud/config` (which is the ccloud tool config file)

clients/cloud/ruby/README.md
edenhill: Add it as a bash step before bundle install, assuming being in repo top dir:
`$ cd clients/ruby`  or whever it is

clients/cloud/ruby/README.md
```diff
@@ -0,0 +1,68 @@
...
+    bootstrap.servers=<broker-1,broker-2,broker-3>
+    sasl.username=<api-key-id>
+    sasl.password=<secret-access-key>
+    ```
+* Install gems
+    ```bash
+    $ bundle install
+    Resolving dependencies...
```
edenhill: You don't need to list the output of the command here.

clients/cloud/ruby/README.md
```diff
@@ -0,0 +1,68 @@
...
+    Producing record: alice	{"count":2}
+    Producing record: alice	{"count":3}
+    Producing record: alice	{"count":4}
+    Producing record: alice	{"count":5}
+    Producing record: alice	{"count":6}
+    Producing record: alice	{"count":7}
+    Producing record: alice	{"count":8}
+    Producing record: alice	{"count":9}
+    Producing record: alice	{"count":10}
+    10 messages were successfully produced to topic test1!
```
edenhill, line 43: I count 11 messages ;)

clients/cloud/ruby/c_cloud.rb
```diff
@@ -0,0 +1,112 @@
...
+      # Trigger a delivery once 5 messages have been buffered.
+      delivery_threshold: 5,
+
+      # Trigger a delivery every 1 second.
```
edenhill: can we do better, e.g 1-10ms?

People will cut'n'paste these examples, change them, and eventually be annoyed that ccloud latency is horrible.

edenhill (changes requested):

clients/cloud/ruby/README.md
stanislavkozlovski: Fair enough. I opted for `ccloud.config` because of this client not being based on librdkafka. This is better for consistency across the examples as well

clients/cloud/ruby/README.md
stanislavkozlovski: This readme is located in `clients/ruby`. Wouldn't users read it from there?

clients/cloud/ruby/README.md
stanislavkozlovski: Good catch, this and the consumer example were outdated - I had fixed it in the code but not here hehe

clients/cloud/ruby/c_cloud.rb
stanislavkozlovski: For sure. It's worth noting that the library gives an example of 30 **seconds**

clients/cloud/ruby/README.md
edenhill: Oh, right, librdkafka.config is indeed a bad name, but so is ccloud.config.  Maybe example.config ?

clients/cloud/ruby/c_cloud.rb
edenhill: Now that is a crazy default latency

clients/cloud/ruby/README.md
```diff
@@ -9,7 +9,7 @@ Produce messages to and consume messages from [Confluent Cloud](https://www.conf
 * Access to a [Confluent Cloud](https://www.confluent.io/confluent-cloud/) cluster
 * Local file with configuration parameters to connect to your Confluent Cloud instance ([how do I find those?](https://docs.confluent.io/current/cloud/using/config-client.html#librdkafka-based-c-clients)). Format the file as follows:
     ```bash
-    $ cat ~/.ccloud/librdkafka.config
+    $ cat ~/.ccloud/example.config
```
edenhill, line 12: @ybyzek Maybe we should use this filename in all of the examples? The files are identical anyway, and the example filename avoids it getting confused with any other config files in ccloud/

stanislavkozlovski: I also think so. I pushed a commit to do that - cc @ybyzek

clients/cloud/ruby/README.md
edenhill: Maybe, but being explicit here doesn't hurt. People tend to need complete instructions

stanislavkozlovski:
@edenhill  @ybyzek is this good to merge?

ybyzek:
@stanislavkozlovski : LGTM but note there are some conflicts in the `README.md`'s because I modified those in another commit too.

stanislavkozlovski:
Rebased 👍

clients/cloud/ruby/c_cloud.rb
```diff
@@ -0,0 +1,112 @@
...
+      ssl_ca_certs_from_system: true
+    )
+  end
+
+  def topic
+    @args[:topic]
+  end
+
+  #
+  # An asynchronous producer tweaked for the low-throughput needs of this example
```
edenhill: I think you can remove the "tweaked .*" part, it is the ruby defaults that are terribly off.

clients/cloud/ruby/producer.rb
```diff
@@ -0,0 +1,60 @@
...
+    begin
+      ccloud.producer.produce(record_value, key: record_key, topic: topic)
+      produced_messages += 1
+    rescue Kafka::Error => e
+      puts "Failed to produce record #{record}: #{e.message}"
+    end
+  end
+ensure
+  # deliver all the buffered messages
+  ccloud.producer.deliver_messages
```
edenhill, line 54: how are delivery errors reported?

stanislavkozlovski: Seems like they're just dropped. The `.shutdown` method actually ensures delivery. Digging into it, it just waits for a Worker thread to shut down when it gives it a `:shutdown` message. Here is how it is handled:
```ruby
when :shutdown
            begin
              # Deliver any pending messages first.
              @producer.deliver_messages
            rescue Error => e
              @logger.error("Failed to deliver messages during shutdown: #{e.message}")

              @instrumenter.instrument("drop_messages.async_producer", {
                message_count: @producer.buffer_size + @queue.size,
              })
            end

            # Stop the run loop.
            break
```
`Error` here is any ruby error

edenhill: So the application would need to check that instrumenter message_count thing to know if messages were dropped?
Isn't there a per-message report/future/whatever?

stanislavkozlovski: Nope
```
If you use the asynchronous producer there's no guarantee that messages will have been delivered after #deliver_messages returns. A way of blocking until a message has been delivered with the asynchronous producer may be implemented in the future.
```

stanislavkozlovski:
Oops, I apparently forgot to push the `tweaked` comment removal commit. Sorry about that.

@edenhill is this good to go?

clients/cloud/ruby/producer.rb
edenhill: That's terrible.

edenhill (approved):
LGTM!

ybyzek:
@stanislavkozlovski , sorry for not noticing this before, but can you please change the destination branch from `master` to `5.0.1-post`?

ybyzek:
I am closing this PR because it was merged to the correct branch with https://github.com/confluentinc/examples/pull/99

PR #77, opened 2018-12-09
DEVX-566: Producer/consumer example to Confluent Cloud: Confluent CLI

ybyzek:


clients/cloud/confluent-cli/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+Produce messages to and consume messages from [Confluent Cloud](https://www.confluent.io/confluent-cloud/) using [Confluent CLI](https://docs.confluent.io/current/cli/index.html).
+
+*Note: The Confluent CLI is meant for development purposes only and is not suitable for a production environment*
+
+
+# Prerequisites
+
+* [Confluent CLI](https://docs.confluent.io/current/cli/installing.html) installed on your machine. It is provided as part of the [Confluent Platform download](https://www.confluent.io/download/).
+* [Confluent Cloud CLI](https://docs.confluent.io/current/cloud/cli/install.html) installed on your machine. It is provided as part of the Confluent Platform package or may be installed separately.
```
edenhill: separately how? brew? go get?

clients/cloud/confluent-cli/README.md
```diff
@@ -0,0 +1,60 @@
...
+$ ccloud topic create test1
+```
+
+2. Run the Confluent CLI producer, passing in additional arguments to parse the message key.
+
+```bash
+$ confluent produce test1 --cloud --property parse.key=true --property key.separator=,
+```
+
+At the `>` prompt, type a few messages, using a `,` as the separator between the message key and value:
```
edenhill, line 37: Having people insert JSON is error prone (in the producer, but even more so in the consumer), but I guess we don't have much choice here.

clients/cloud/confluent-cli/confluent-cli-example.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+topic_name=test1
+
+ccloud topic create $topic_name
+
+num_messages=10
+for ((i=0;i<$num_messages;i++)); do
+  echo "alice,{\"count\":${i}}" | confluent produce $topic_name --cloud --property parse.key=true --property key.separator=,
```
edenhill: this instantiates a client per message, which is expensive, wrap the loop as so instead:
```bash
(for ((i=0;i<$num_messages;i++)); do
  echo "alice,{\"count\":${i}}" ;
done) | confluent produce $topic_name --cloud --property parse.key=true --property key.separator=,
```

clients/cloud/confluent-cli/confluent-cli-example.sh
```diff
@@ -0,0 +1,12 @@
...
+done
+
+confluent consume $topic_name --cloud --property print.key=true --from-beginning --max-messages $num_messages
```
edenhill: If the example is re-run it will still only show the messages from the first run.
Suggest removing --max-messages, user can always ctrl-c.

edenhill (changes requested):

clients/cloud/confluent-cli/README.md
ybyzek: Added link

clients/cloud/confluent-cli/README.md
ybyzek: @edenhill : for those in favor of consistency between the examples, we need to accept this for now IMO

clients/cloud/confluent-cli/confluent-cli-example.sh
ybyzek: @edenhill genius idea, thank you!

clients/cloud/confluent-cli/confluent-cli-example.sh
ybyzek: Agree!

clients/cloud/confluent-cli/confluent-cli-example.sh
edenhill: this can be simplified even further by: `( for i in seq 0 10 ; do echo "alice,{...$i..}" ; done) | confluent produce ..`

ybyzek: I refactored this a touch (for printing nicely), but kept the num messages as a var.

clients/cloud/confluent-cli/confluent-cli-example.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Set topic name
+topic_name=test1
+
+# Create topic in Confluent Cloud
+ccloud topic create $topic_name
+
+# Produce messages
+num_messages=10
+(for ((i=0;i<$num_messages;i++)); do echo "alice,{\"count\":${i}}" ; done) | \
```
edenhill: Yeah, I mean you can use `for i in seq 0 $num_messages; do ..` to make it simpler. Anyway, doesn't matter

clients/cloud/confluent-cli/confluent-cli-example.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
```
edenhill, line 2: `set -eu` is good practice

edenhill (approved):
LGTM!
two minor nits remaining, do with them as you please

clients/cloud/confluent-cli/confluent-cli-example.sh
ybyzek: I like it!

clients/cloud/confluent-cli/confluent-cli-example.sh
ybyzek: Simplified a bit

PR #78, opened 2018-12-09
DEVX-555, DEVX-547: Make Java producer/consumer & Kafka Streams examples to Confluent Cloud

ybyzek:


clients/cloud/java/README.md
```diff
@@ -1,95 +1,118 @@
...
 	```shell
-	# Create page_visits topic
-	$ ccloud topic create page_visits
-	```
-	You should see
+	# Build the client examples
+	$ mvn clean package
 	
+	# Run the producer
+	$ mvn exec:java -Dexec.mainClass="io.confluent.examples.clients.ProducerExample" \
```
edenhill: should there be a "cloud" somewhere in that namespace?

clients/cloud/java/README.md
```diff
@@ -1,95 +1,118 @@
...
+
+2. Run the consumer, passing in arguments for (a) the local file with configuration parameters to connect to your Confluent Cloud instance and (b) the same topic name as used above. Verify that the consumer received all the messages:
+
+	```shell
+	# Build the client examples
+	$ mvn clean package
 	
-	You should see
+	# Run the consumer
+	$ mvn exec:java -Dexec.mainClass="io.confluent.examples.clients.ConsumerExample" \
```
edenhill: dito

clients/cloud/java/README.md
```diff
@@ -1,95 +1,118 @@
...
+        When you are done, press `<ctrl>-c`.
+
+3. Run the Kafka Streams application, , passing in arguments for (a) the local file with configuration parameters to connect to your Confluent Cloud instance and (b) the same topic name as used above. Verify that the consumer received all the messages:
+
+        ```shell
+        # Build the client examples
+        $ mvn clean package
+
+        # Run the consumer
+        $ mvn exec:java -Dexec.mainClass="io.confluent.examples.clients.StreamsExample" \
```
edenhill: dito

clients/cloud/java/pom.xml
```diff
@@ -35,9 +35,9 @@
 
     <properties>
         <java.version>1.8</java.version>
-        <kafka.version>2.0.2-SNAPSHOT</kafka.version>
+        <kafka.version>2.1.0</kafka.version>
         <kafka.scala.version>2.11</kafka.scala.version>
-        <confluent.version>5.0.2-SNAPSHOT</confluent.version>
+        <confluent.version>5.0.1</confluent.version>
```
edenhill, line 40: How will we remember to update these as 5.1 is released?

clients/cloud/java/src/main/java/io/confluent/examples/clients/ConsumerExample.java
```diff
@@ -42,23 +49,30 @@ public static void main(String[] args) throws Exception {
...
     // Add additional properties.
-    props.put("enable.auto.commit", "true");
-    props.put("auto.commit.interval.ms", "1000");
-    props.put("key.deserializer", "org.apache.kafka.common.serialization.StringDeserializer");
-    props.put("value.deserializer", "org.apache.kafka.common.serialization.StringDeserializer");
-
-    // WARNING: These settings are for CLI tools / examples. Don't use in production unless you want to reset the consumer app every reboot.
-    props.put("group.id", "example" + new Random().nextInt());
-    props.put("auto.offset.reset", "earliest");
+    props.put(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, "true");
```
edenhill: this is the default value

clients/cloud/java/src/main/java/io/confluent/examples/clients/ConsumerExample.java
```diff
@@ -42,23 +49,30 @@ public static void main(String[] args) throws Exception {
...
+    props.put(ConsumerConfig.AUTO_COMMIT_INTERVAL_MS_CONFIG, "1000");
```
edenhill: since this application runs for such a short time even this value probably won't be used, but commits will be performed when closing the consumer, so I don't think this config is really needed.

clients/cloud/java/src/main/java/io/confluent/examples/clients/ConsumerExample.java
```diff
@@ -42,23 +49,30 @@ public static void main(String[] args) throws Exception {
...
+    props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, "org.apache.kafka.connect.json.JsonDeserializer");
+    props.put(ConsumerConfig.GROUP_ID_CONFIG, "java_example_group_1");
+    props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
 
-    final Consumer<String, String> consumer = new KafkaConsumer<String, String>(props);
+    final Consumer<String, JsonNode> consumer = new KafkaConsumer<String, JsonNode>(props);
     consumer.subscribe(Arrays.asList(topic));
 
+    Long total_count = 0L;
+    String key;
```
edenhill: key and value should be in the `while (true)` loop (only scope where they are used)

clients/cloud/java/src/main/java/io/confluent/examples/clients/ConsumerExample.java
```diff
@@ -42,23 +49,30 @@ public static void main(String[] args) throws Exception {
...
     try {
       while (true) {
-        ConsumerRecords<String, String> records = consumer.poll(100);
-        for (ConsumerRecord<String, String> record : records) {
-          System.out.printf("offset = %d, key = %s, value = %s \n", record.offset(), record.key(), record.value());
+        ConsumerRecords<String, JsonNode> records = consumer.poll(100);
+        for (ConsumerRecord<String, JsonNode> record : records) {
+          key = record.key();
+          value = record.value();
+          countRecord = MAPPER.treeToValue(value,RecordJSON.class);
```
edenhill: missing space after comma here: `value,RecordJSON..`

clients/cloud/java/src/main/java/io/confluent/examples/clients/ProducerExample.java
```diff
@@ -16,57 +16,86 @@
...
-    props.put("batch.size", 16384);
-    props.put("linger.ms", 100);
-    props.put("compression.type", "lz4");
-    props.put("buffer.memory", 33554432);
-    props.put("key.serializer", "org.apache.kafka.common.serialization.StringSerializer");
-    props.put("value.serializer", "org.apache.kafka.common.serialization.StringSerializer");
-
+    // Create topic if needed
+    String topic = args[1];
+    createTopic(topic, 12, 3, props);
```
edenhill: 12 partitions? the other examples used one partition

clients/cloud/java/src/main/java/io/confluent/examples/clients/ProducerExample.java
```diff
@@ -16,57 +16,86 @@
...
 
-    Producer<String, String> producer = new KafkaProducer<String, String>(props);
+    // Add additional properties.
+    props.put(ProducerConfig.ACKS_CONFIG, "all");
+    props.put(ProducerConfig.RETRIES_CONFIG, 200);
+    props.put(ProducerConfig.BATCH_SIZE_CONFIG, 16384);
+    props.put(ProducerConfig.LINGER_MS_CONFIG, 100);
+    props.put(ProducerConfig.BUFFER_MEMORY_CONFIG, 33554432);
```
edenhill: I don't think retries, batch size or buffer memory config is needed, we're not doing anything special.

clients/cloud/java/src/main/java/io/confluent/examples/clients/ProducerExample.java
```diff
@@ -16,57 +16,86 @@
...
-      String record = String.format("%s,%s,%s", runtime, site, ip);
+    // Produce sample data
+    final Long numMessages = 10L;
+    for (Long i = 0L; i < numMessages; i++) {
+      String key = "alice";
+      RecordJSON countRecord = new RecordJSON(i);
+      JsonNode value = MAPPER.valueToTree(countRecord);
 
-      producer.send(new ProducerRecord<>(topic, ip, record));
+      producer.send(new ProducerRecord<String, JsonNode>(topic, key, value));
```
edenhill: these send()s are asynchronous, but there is nothing checking if they succeeded

clients/cloud/java/src/main/java/io/confluent/examples/clients/ProducerExample.java
```diff
@@ -16,57 +16,86 @@
...
+      System.out.printf("Produced record: %s\t%s%n", key, value);
     }
 
-    System.out.printf("Successfully produced %s messages to %s.\n", events, topic);
+    System.out.printf("10 messages were successfully produced to topic %s%n", topic);
```
edenhill: since the send was asynch we don't know if the messages were successfully produced.
Suggesting using a callback to send() for error reporting, and calling flush() here to wait for all messages to deliver/fail.

edenhill (changes requested):

clients/cloud/java/pom.xml
ybyzek: DevX is responsible for updating the release for the entire `examples` repo.

clients/cloud/java/src/main/java/io/confluent/examples/clients/ConsumerExample.java
ybyzek: Removed

clients/cloud/java/src/main/java/io/confluent/examples/clients/ConsumerExample.java
ybyzek: I declared them outside the while loop purposely to avoid instantiating a new object each time, but my understanding may be incorrect and if so, then I'm happy to move inside the loop.

clients/cloud/java/src/main/java/io/confluent/examples/clients/ProducerExample.java
ybyzek: Agree!

clients/cloud/java/src/main/java/io/confluent/examples/clients/ConsumerExample.java
ybyzek: After some Googling: I suppose today's compilers are smart enough to handle this.  I'll move it in.

clients/cloud/java/src/main/java/io/confluent/examples/clients/ConsumerExample.java
edenhill: reminder to remove this line.

clients/cloud/java/README.md
ybyzek: Ack to this and all references

clients/cloud/java/src/main/java/io/confluent/examples/clients/ProducerExample.java
ybyzek: Ack

clients/cloud/java/src/main/java/io/confluent/examples/clients/ConsumerExample.java
ybyzek: K

clients/cloud/java/src/main/java/io/confluent/examples/clients/ProducerExample.java
ybyzek: Implemented Callback

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/ProducerExample.java
```diff
@@ -0,0 +1,122 @@
+/**
+ * Copyright 2015 Confluent Inc.
```
edenhill: 2018, ish

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/ProducerExample.java
```diff
@@ -0,0 +1,122 @@
...
+          if (!(e.getCause() instanceof TopicExistsException)) {
+              throw new RuntimeException(e);
+          }
+      }
+  }
+
+  public static void main(String[] args) throws IOException {
+    if (args.length != 2) {
+      System.out.println("Please provide command line arguments: configPath topic");
+      System.exit(-1);
```
edenhill: exit values should be positive, 1 is typical for error .

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/ProducerExample.java
```diff
@@ -0,0 +1,122 @@
...
+            } else {
+              System.out.printf("Produced record to topic %s partition [%d] @ offset %d%n", m.topic(), m.partition(), m.offset());
+            }
+          }
+      });
+    }
+
+    producer.flush();
+
+    System.out.printf("10 messages were successfully produced to topic %s%n", topic);
```
edenhill: Since onCompletion neither fails or counts errors, this printout could be a lie.

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/StreamsExample.java
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright 2017 Confluent Inc.
```
edenhill: closer, but still not quite there

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/StreamsExample.java
```diff
@@ -0,0 +1,102 @@
...
+          System.out.println("Please provide command line arguments: configPath topic");
+          System.exit(-1);
+        }
+    
+        String topic = args[1];
+    
+        // Load properties from disk.
+        Properties props = loadConfig(args[0]);
+        props.put(StreamsConfig.APPLICATION_ID_CONFIG, "java_streams_example_group_1");
+        props.put(StreamsConfig.COMMIT_INTERVAL_MS_CONFIG, 1000);
```
edenhill: no need for this

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/StreamsExample.java
```diff
@@ -0,0 +1,102 @@
...
+        props.put(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG, 0);
```
edenhill, line 68: Explain with a short comment why this is needed.

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/StreamsExample.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public class StreamsExample {
+
+    public static void main(String[] args) throws Exception {
+
+        if (args.length != 2) {
+          System.out.println("Please provide command line arguments: configPath topic");
+          System.exit(-1);
```
edenhill: exit(1)

clients/cloud/python/README.md
```diff
@@ -50,7 +50,7 @@ Successfully produced record to topic test1 partition [0] @ offset 9
 10 messages were successfully produced to topic test1!
 ```
 
-1. Run the consumer, passing in arguments for (a) the local file with configuration parameters to connect to your Confluent Cloud instance and (b) the same topic name as used above. Verify that the consumer received all the messages:
+2. Run the consumer, passing in arguments for (a) the local file with configuration parameters to connect to your Confluent Cloud instance and (b) the same topic name as used above. Verify that the consumer received all the messages:
 
 ```bash
 $ ./consumer.py -f ~/.ccloud/librdkafka.config -t test1
```
edenhill: we're changing this to example.config, right? (maybe in another PR?)

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/ConsumerExample.java
```diff
@@ -13,12 +13,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.confluent.examples.clients;
```
edenhill, line 16: ^ copyright year somewhere up there

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/ConsumerExample.java
```diff
@@ -27,9 +32,11 @@
 import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.Properties;
-import java.util.Random;
 
 public class ConsumerExample {
+
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
```
edenhill: change to exit(1) somewhere down there ---v

edenhill (changes requested):

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/ProducerExample.java
ybyzek: Changed to 1.

These are all errors resulting from copy/paste other code (from 2015, apparently)

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/ProducerExample.java
ybyzek: This is actually the same message in all our code examples, even the python one: 

https://github.com/confluentinc/examples/blob/5.0.1-post/clients/cloud/python/producer.py#L96

I understand your point, but IMO it's ok.  As long as the callback prints the error if there is one (which the callback does)

edenhill: You're asking me to be consistent in my reviews?!

All we need to do is remove "successfully" for this printout to be true, regardless of delivery status.

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/StreamsExample.java
ybyzek: Removed

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/StreamsExample.java
ybyzek: Done

clients/cloud/python/README.md
ybyzek: I didn't for the Java example, it can (should?) use the one established by Confluent Cloud CLI?  WDYT?

edenhill: Yeah, the Java clients can/should use ~/.ccloud/config directly.
But since there is java-specific config in there it can't be reused for the other clients, it does help if all the other clients share the same config file though.

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/ProducerExample.java
ybyzek: Removed successfully.

clients/cloud/python/README.md
ybyzek: Agree.  Non-java clients should use `~/.ccloud/example.config`

vvcephei (approved):
Hi @ybyzek ,

I've made a pass on this, and I really like it.

The readme guide is very clear, and has a good flow.

The code is also quite readable and it all looks good to me.

The only question I have is around the relationship among the consumer's serializer, the producer's deserializer, and the streams serde.
* consumer: org.apache.kafka.connect.json.JsonDeserializer
* producer: org.apache.kafka.connect.json.JsonSerializer
* streams: io.confluent.examples.clients.cloud.JsonPOJOSerde

It seems asymmetrical to me to mix the usage of Connect's serializer/deserializer with our own serde here. In principle, JSON is a lingua franca, so it's fine to use different serdes, but as a recommendation or usage example, I'm not so sure.

This might also be a shortcoming of the meta-design of the whole system, in which case, perhaps we should try to improve the design of the system wrt serdes.

ybyzek:
@vvcephei : I 💯 agree with your comment and prefer to showcase a consistent JSON serdes usage.  The reason I wrote the code the way I did was only one (lame) reason: I was pulling from examples I could find and it was all that I could get working.

So stepping back, what is the right way to use JSON in all three examples (Producer, Consumer, Streams)?

vvcephei:
@ybyzek ,

I think your experience with this is a great proxy for a "developer's experience" ;) . Thanks for doing this!

AFAIK, we don't have any pre-packaged notion of a serde for transforming between JSON and POJOs, so I think anyone who wants it would have to provide their own, as you've done.

I suppose the consistent approach would then be to grab the serializer and deserializer from your serde, and then construct the KafkaProducer and KafkaConsumer with the constructors that accept instances of Serializer and Deserializer, respectively. Does that make sense?

ybyzek:
@vvcephei the other option is to use the String Serdes instead of JSON object, and just write out the json object as a String.  This may be more straight-forward.  The question is which is more common?

ybyzek:
This is the alternative I mentioned: String Serdes instead of JSON Serdes: 

https://github.com/confluentinc/examples/pull/82

The Streams example still uses POJO though...not sure if there's an easier way in kafka streams API to handle that

vvcephei:
I think the String serdes would be a bit of a cop-out, supposing people want to use JSON, having String serdes doesn't really suit the use case.

ybyzek:
@vvcephei : now updated to use the Confluent utilities for json serdes (many thanks to @mikebin )

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/ConsumerExample.java
```diff
@@ -42,23 +45,25 @@ public static void main(String[] args) throws Exception {
...
     // Add additional properties.
-    props.put("enable.auto.commit", "true");
-    props.put("auto.commit.interval.ms", "1000");
-    props.put("key.deserializer", "org.apache.kafka.common.serialization.StringDeserializer");
-    props.put("value.deserializer", "org.apache.kafka.common.serialization.StringDeserializer");
+    props.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, "org.apache.kafka.common.serialization.StringDeserializer");
+    props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, "io.confluent.kafka.serializers.KafkaJsonDeserializer");
+    props.put(KafkaJsonDeserializerConfig.JSON_VALUE_TYPE, DataRecord.class);
+    props.put(ConsumerConfig.GROUP_ID_CONFIG, "java_example_group_1");
+    props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
```
hachikuji, line 52: Just double-checking that the intent is to override any user-provided configurations.

hachikuji (commented):
Looks good. Just had one question.

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/ConsumerExample.java
ybyzek: @hachikuji : by `user-provided configurations`, do you mean those provided in the CCloud config file that is passed into the app, or something else?

edenhill: @hachikuji I think the idea here is to show-case an example application. The application requires auto.offset.reset=earliest and the example group id (to avoid conflict with proper groups), so I thin kit is fine to override whatever was configured in .ccloud/config for these properties.

hachikuji: @ybyzek Yeah, that's what I was referring to. I guess we are mainly using that config to provide connection-related configs.

@edenhill Yep, seems reasonable to me. My brain is programmed to notice these things, but it seems sensible for the example.

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/ProducerExample.java
```diff
@@ -0,0 +1,117 @@
...
+            } else {
+              System.out.printf("Produced record to topic %s partition [%d] @ offset %d%n", m.topic(), m.partition(), m.offset());
+            }
+          }
+      });
+    }
+
+    producer.flush();
+
+    System.out.printf("10 messages were produced to topic %s%n", topic);
```
hachikuji, line 101: nit: probably fine for the example, but technically some of the produce attempts may have failed.

hachikuji (approved):
One minor nitpick, but otherwise LGTM.

clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/ProducerExample.java
ybyzek: @edenhill also had made a comment along those lines when the previous verbiage included the word `successfully` -- which has since been removed :)  I'm going to leave the example as-is, we can revisit later if needed

PR #96, opened 2019-01-08
DOCS-1153: catch SerializationException

ybyzek:
To setup for tests that disable auto schema registration


clients/avro/src/main/java/io/confluent/examples/clients/basicavro/ProducerExample.java
```diff
@@ -35,13 +36,14 @@ public static void main(final String[] args) {
                 Thread.sleep(1000L);
             }
 
+            System.out.printf("Successfully produced 10 messages to a topic called %s%n", TOPIC);
```
edenhill, line 40: There should be a flush above this line.

edenhill (changes requested):

clients/avro/src/main/java/io/confluent/examples/clients/basicavro/ProducerExample.java
ybyzek: Ack and done

edenhill (approved):
LGTM

PR #103, opened 2019-01-15
DEVX-548: Producer/consumer example for Confluent Cloud: Node.js

colinhicks:
- [x] add readme boiler plate and link top-level readme
- [x] fix producer delivery reports
- [x] make consumer output consistent with canonical example
- [x] license headers


clients/cloud/nodejs/config.js
```diff
@@ -0,0 +1,99 @@
+const minimist = require('minimist');
```
edenhill, line 17: add license header

clients/cloud/nodejs/consumer.js
```diff
@@ -0,0 +1,42 @@
...
+
+  consumer.subscribe([config.topic]);
+  consumer.consume();
+}
+
+consumerExample()
+  .catch((err) => {
+    console.error(`Something went wrong:\n${err}`);
+    process.exit(1);
+  });
```
edenhill: missing newline at end of file.

ybyzek:
@colinhicks nice!  Thanks for the PR!

I've copied the clients team for review of the code.

Meanwhile, could you please provide corresponding README?

1. One for Node.js (similar to what is done for each language, e.g. here is the [Python one](https://github.com/confluentinc/examples/blob/5.1.0-post/clients/cloud/python/README.md))
2. Modify the [higher-level README](https://github.com/confluentinc/examples/blob/5.1.0-post/clients/cloud/README.md) to include a link to Node.js

clients/cloud/nodejs/consumer.js
```diff
@@ -0,0 +1,42 @@
...
+
+  const consumer = await createConsumer(config);
+
+  consumer.subscribe([config.topic]);
+  consumer.consume();
+}
+
+consumerExample()
+  .catch((err) => {
+    console.error(`Something went wrong:\n${err}`);
```
edenhill, line 73: If these are consumer errors then they should be logged and ignored, since the client automatically recovers from all errors.

clients/cloud/nodejs/producer.js
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  producer.flush(500, () => {
+    producer.disconnect();
+  });
+}
+
+produceExample()
+  .catch((err) => {
+    console.error(`Something went wrong:\n${err}`);
```
edenhill, line 113: same here, the producer auto recovers

clients/cloud/nodejs/producer.js
```diff
@@ -0,0 +1,57 @@
...
+        console.warn('event.error', err);
+        reject(err);
+    });
+
+    producer.connect();
+  });
+}
+
+function range(length) {
+  return Array.from(Array(length).keys());
```
edenhill: I don't really know nodejs, but is creating two arrays really the most effective way to iterate over 10 numbers?

edenhill (changes requested):
Add license headers to each code file.
Add a README (copy one of the others and update)

clients/cloud/nodejs/producer.js
colinhicks: Yeah, there are different ways to do this for sure. The outer `Array.from` can be avoided if we reuse the iterator returned by `keys()` and consume using a `for...of` loop. Debatably, the usage of the latter is more exotic in JavaScript-land, compared to the `forEach` alternative. Of course there's a good argument for a standard `while` loop, too. 

I'll push the `for...of` option next, so we can feel it out.

clients/cloud/nodejs/consumer.js
colinhicks: This actually doesn't catch consumer errors. It's the top-level error handler for the promise chain. It would catch, for example, a config file access error.

colinhicks:
Thanks for the review! I should have noted in Slack that what's pushed so far is incomplete, with respect to the (great) template. I plan to circle back this evening.

clients/cloud/nodejs/producer.js
```diff
@@ -36,14 +32,16 @@ async function produceExample() {
 
   const producer = await createProducer(config);
 
-  range(10).forEach((count) => {
+  const range = Array(10).keys();
```
edenhill: This is still instantiating an array for iterating over a fixed range of integers, I don't really understand why would you want to do that? Why not just use a for loop?

mhowlett: it's creating an empty array of length 10 and then creating another array of length 10 with the range values in it. this (lunacy) is in fact idiomatic javascript (and so probably optimized away nicely), but i wouldn't do this here since it's then just being used in a for loop. I'd only do this if I wanted to `map` over it or something. writing `for (let i=0; i<10; ++i)` is also terser and more straightforward (IMO).

clients/cloud/nodejs/producer.js
```diff
@@ -0,0 +1,55 @@
...
+  for (const count of range) {
+    const key = 'alice';
+    const value = Buffer.from(JSON.stringify({ count }));
+    
+    console.log(`Producing record ${key}\t${value}`);
+
+    producer.produce(config.topic, -1, value, key);
+  }
+
+  producer.flush(500, () => {
```
mhowlett: 500 doesn't give much margin for error, would increase to maybe 10000.

clients/cloud/nodejs/producer.js
colinhicks: Totally happy to change this detail. My excuse for not using `for` from the start is pre-ES6 JS, when block scoped vars weren't a thing and ... Clojure.

clients/cloud/nodejs/config.js
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+    reader
+      .on('line', (line) => lines.push(line))
+      .on('close', () => resolve(lines));    
+  });
+}
+
+async function configFromPath(path) {
+  await assertRead(path);
```
mhowlett: structurally, the `assertRead` functionality would be best encapsulated inside `readAllLines`. Also, the error event on the read stream would be best used for errors, otherwise there's a race condition on the file being removed.

clients/cloud/nodejs/consumer.js
```diff
@@ -0,0 +1,42 @@
...
+    'sasl.mechanisms': 'PLAIN',
+    'group.id': 'node-example-group-1',
+    'auto.offset.reset': 'earliest'
+  });
+
+  return new Promise((resolve, reject) => {
+    consumer
+      .on('ready', () => resolve(consumer))
+      .on('data', ({topic, offset, partition, key, value}) =>
+        console.log(`Received record from topic ${topic} partition ${partition} @ ${offset}: ${key.toString()}\t${value.toString()}`));
```
mhowlett: I'd pass a `messageHandler` function as a parameter into `createConsumer`, call that here and specify the `console.log` behavior in the `consumerExample` method. I know it's only used once, but it seems overly constrained otherwise, given you've gone to the trouble of making the abstraction (which I think is good and appropriate btw).

clients/cloud/nodejs/producer.js
```diff
@@ -0,0 +1,55 @@
...
+const { configFromCli } = require('./config');
+
+async function createProducer(config) {
+  const producer = new Kafka.Producer({
+    'metadata.broker.list': config['bootstrap.servers'],
+    'sasl.username': config['sasl.username'],
+    'sasl.password': config['sasl.password'],
+    'security.protocol': 'SASL_SSL',
+    'sasl.mechanisms': 'PLAIN',
+    'dr_cb': true
```
mhowlett: you're turning delivery report events on here, but not subscribing to them. I notice in the python example, there is a handler, which prints the success or otherwise, and I think it makes sense to follow that here. https://github.com/confluentinc/examples/blob/5.1.0-post/clients/cloud/python/producer.py#L79

like the consumer, i'd make that a parameter of the `createProducer` function and define the behavior in `produceExample`.

clients/cloud/nodejs/consumer.js
colinhicks: Good suggestion. I realized I can achieve a similar effect by passing a callback to `consume`. This is my first experience with node-rdkafka. I've been inclined to gently pave over some of its opinions, but I'm still feeling out the balance.

clients/cloud/nodejs/config.js
colinhicks: There's some extra logic here that we could ideally remove, which would also remove the appearance of a race condition.

Bit of background: `assertRead` is redundant because the stream reader would raise an exception on EOENT, if the former check were removed. However, the reader yields such an exception outside the scope of `readAllLines`, somewhere in the `lazyFs` impl. I haven't yet found a feasible way to catch this, so I added the redundant assertRead to fail fast & more gracefully, within the promise chain.

clients/cloud/nodejs/producer.js
colinhicks: Yep. I had enabled `db_cb` pursuant to the TODO above.

clients/cloud/nodejs/config.js
colinhicks: Update: I've moved the file access assertion logic within `readAllLines`.

clients/cloud/nodejs/README.md
```diff
@@ -0,0 +1,74 @@
...
+* [OpenSSL](https://www.openssl.org) version 1.0.2.
+* Access to a [Confluent Cloud](https://www.confluent.io/confluent-cloud/) cluster
+* Local file with configuration parameters to connect to your Confluent Cloud instance ([how do I find those?](https://docs.confluent.io/current/cloud/using/config-client.html#librdkafka-based-c-clients)). Format the file as follows:
+```bash
+$ cat ~/.ccloud/example.config
+bootstrap.servers=<broker-1,broker-2,broker-3>
+sasl.username=<api-key-id>
+sasl.password=<secret-access-key>
+```
+* Users of macOS 10.13 (High Sierra) and above may need to specify OpenSSL paths so that the client's [librdkafka](https://github.com/edenhill/librdkafka) bindings can be built correctly:
```
ybyzek: @colinhicks is this requirement unique to Node.js or all clients?

ybyzek: Actually, perhaps all of this can be cut from this README, and instead provide a pointer to https://github.com/Blizzard/node-rdkafka#requirements ?

colinhicks: Sure, we could link the client documentation instead, @ybyzek.

I think it's okay to consider this setup step scoped to node-rdkafka specifically, in that it's the mechanism that library uses to override the default linkage, here: https://github.com/Blizzard/node-rdkafka/blob/64fdf8daccd5492ae39a3e6119b8404c52232000/binding.gyp#L134-L144. 

More generally, my sense is that one of the following applies to this issue, across librdkafka libs:
* The library requires a similar setup step, and we should expect it to be part of that library's documentation. (this case)
* The library handles this issue transparently. (e.g. https://github.com/appsignal/rdkafka-ruby/blob/4c320b4df963a6118f45f3910042cffd685cdd1d/ext/Rakefile#L26-L29)
* The library doesn't yet handle it, but given any scale of user base, the issue will be raised.

clients/cloud/nodejs/producer.js
```diff
@@ -0,0 +1,116 @@
...
+
+function createProducer(config, onDeliveryReport) {
+  const producer = new Kafka.Producer({
+    'metadata.broker.list': config['bootstrap.servers'],
+    'sasl.username': config['sasl.username'],
+    'sasl.password': config['sasl.password'],
+    'security.protocol': 'SASL_SSL',
+    'sasl.mechanisms': 'PLAIN',
+    'dr_cb': true,
+    'dr_msg_cb': true
```
mhowlett, line 64: I think you can remove `dr_cb` - from the README docs it seems it's unnecessary given `dr_msg_cb` is specified.

mhowlett (approved):
One more minor comment then lgtm

clients/cloud/nodejs/producer.js
colinhicks: I found that this was necessary in order to populate `value` in the delivery report object. We could remove the `value` from the message logged to the user and also remove this property, if that's advisable.

clients/cloud/nodejs/README.md
colinhicks: @ybyzek Sorry, I just realized that you suggested a pointer to the more general `#requirements` anchor. I made the previous comment and README change w/r/t the OpenSSL issue. And I missed your point that a bigger chunk of the reqs could be delegated to node-rdkafka, although we do require a more recent Node.js version in our example.

Let's just do whatever you think strikes the best balance :).

ybyzek: @colinhicks generally speaking I prefer to de-dupe.  If there is another page that has the full set of install requirements/steps, then I prefer to link to that instead of repeating it.

So do you think https://github.com/Blizzard/node-rdkafka#requirements or https://github.com/Blizzard/node-rdkafka/blob/56c31c4e81f2a042666160338ad65dc4f8f2d87e/README.md#mac-os-high-sierra--mojave is complete?  If yes, I prefer to just link to one of those pages, and from this README we'd remove the `npm install`, etc commands.

colinhicks: Ah, my answer to that question is a qualified no. I think the current readme is actually reasonable and not unlike other node examples in the wild.

In particular the `npm install` portion is important to keep in situ. It's similar in pattern to the Ruby example's Bundler instructions.

The original rub that led us here is the macOS/OpenSSL binding issue. It's an unfortunate wart. But I think it's fine to cover it with the link as-is.

colinhicks:
@edenhill do the changes since your review look good?

clients/cloud/nodejs/README.md
ybyzek: @colinhicks makes sense.  In that case, for your consideration, move the two bullet points about `npm install` and `macOS` to be higher up in the bullet list because they are more closely related to the Node.js install (could even be indented bullets?).  Last item in the list should be about the ccloud config.

ybyzek (approved):
@colinhicks thanks for the PR!  I left one final comment about the README related to bullet point re-ordering, and after that, LGTM!

ybyzek:
@colinhicks , @edenhill : is this good to merge?

clients/cloud/nodejs/config.js
```diff
@@ -0,0 +1,114 @@
...
+
+const title = "Example Node.js Confluent Cloud client";
+
+const requiredOpts = [
+  ['config', '--config CONFIG', 'The path to your Confluent Cloud configuration file'], 
+  ['topic', '--topic TOPIC', 'The topic name on which to operate']
+];
+
+const requiredConfig = [
+  ['bootstrap.servers', 'bootstrap.servers=<URIs>', 'One or more Kafka bootstrap server addresses. Separate multiple with commas.'],
```
edenhill: nit: the bootstrap servers are not URIs (does not have the scheme: prefix), but simple CSV list of host[:port].

edenhill: Help should probably say "Your Confluent Cloud cluster bootstrap servers..." to be consistent with sasl.username/pass below.

clients/cloud/nodejs/config.js
```diff
@@ -0,0 +1,114 @@
...
+function readAllLines(path) {    
+  return new Promise((resolve, reject) => {
+    // Test file access directly, so that we can fail fast.
+    // Otherwise, an ENOENT is thrown in the global scope by the readline internals.
+    try {
+      fs.accessSync(path, fs.constants.R_OK);
+    } catch (err) {
+      reject(err);
+    }
+
```
edenhill, line 88: nit: accessSync() + later createReadStream() actually creates a race condition if the file perms are changed, or file removed, between those two points, so it is generally not recommended, but I don't believe it matters in this case.

clients/cloud/nodejs/consumer.js
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * =============================================================================
+ */
+
+const Kafka = require('node-rdkafka');
+const { configFromCli } = require('./config');
+
+function createConsumer(config, onData) {
+  const consumer = new Kafka.KafkaConsumer({
+    'metadata.broker.list': config['bootstrap.servers'],
```
edenhill: use 'bootstrap.servers' for consistency. 'metadata.broker.list' is the legacy property name.

clients/cloud/nodejs/producer.js
```diff
@@ -0,0 +1,115 @@
...
+ */ 
+
+const Kafka = require('node-rdkafka');
+const { configFromCli } = require('./config');
+
+const ERR_TOPIC_ALREADY_EXISTS = 36;
+
+function ensureTopicExists(config) {
+  const adminClient = Kafka.AdminClient.create({
+    'metadata.broker.list': config['bootstrap.servers'],
```
edenhill: bootstrap.servers

clients/cloud/nodejs/producer.js
```diff
@@ -0,0 +1,115 @@
...
+      }
+
+      return reject(err);
+    });
+  });
+}
+
+function createProducer(config, onDeliveryReport) {
+  const producer = new Kafka.Producer({
+    'metadata.broker.list': config['bootstrap.servers'],
```
edenhill: bootstrap.servers

edenhill (changes requested):
Looks good, just some minor nits on naming consistency which I think would be good to fix.

clients/cloud/nodejs/config.js
colinhicks: Good point - thanks for the correction and suggestion

clients/cloud/nodejs/config.js
colinhicks: Matt and I had a discussion on this, too: https://github.com/confluentinc/examples/pull/103#discussion_r248014598

I agree it's probably okay like this, in this case.

colinhicks:
Thanks for the further feedback! It's addressed in the latest commit.

Regarding updating `metadata.broker.list` to `bootstrap.servers`, I wonder if we should suggest the same change to the Blizzard folks. cf. https://github.com/Blizzard/node-rdkafka#sending-messages

edenhill:
@colinhicks 
> Regarding updating `metadata.broker.list` to `bootstrap.servers` ..

That's a good idea! Do you want to do it, or shall I?

edenhill (approved):
LGTM!

colinhicks:
@edenhill: I'll submit a PR to node-rdkafka for the `bootstrap.servers` change.

@ybyzek: This is ready for merge.

Thanks again for the reviews, all.

PR #114, opened 2019-02-13
Update ksql-workshop.adoc

ghost:
"Stream created" instead of "Table created"


ybyzek (approved):
@Dawlys good catch and thanks for the PR!

PR #139, opened 2019-03-28
Bumping check style version

gAmUssA:
Github reports that 8.12 contains compromised transitive dependency. Need to upgrade


gAmUssA:
wrt 
image

and 

image

ybyzek (changes requested):
@gAmUssA -- thanks for submitting the PR. Have you run the examples to verify they still work with the version change?  I don't think we should merge this PR until it's verified

gAmUssA:
@ybyzek do we strictly obey checkstyle checks?
Because checks require some code changes

gAmUssA:
@ybyzek with 4903fdf I fixed checkstyle warnings

ybyzek (approved):
LGTM!  Thanks again @gAmUssA

PR #151, opened 2019-04-10
DEVX-800: Add Avro to Python CCloud example and integration with Confluent Cloud Schema Registry

ybyzek:


edenhill:
@rnpridgeon This is your home turf, will you do us the honour of reviewing?

clients/cloud/python/README.md
```diff
@@ -69,3 +69,94 @@ Consumed record with key alice and value {"count": 9}, and updated total count t
 Waiting for message or event/error in poll()
 ...
 ```
+
+
+# Example 2: Avro And Confluent Cloud Schema Registry
```
rnpridgeon, line 74: Can we add a note about the regression introduced when not specifying the root CA.

clients/cloud/python/consumer_ccsr.py
```diff
@@ -0,0 +1,94 @@
...
+    #   topic if no committed offsets exist
+    c = AvroConsumer({
+        'bootstrap.servers': conf['bootstrap.servers'],
+        'sasl.mechanisms': 'PLAIN',
+        'security.protocol': 'SASL_SSL',
+        'sasl.username': conf['sasl.username'],
+        'sasl.password': conf['sasl.password'],
+        'schema.registry.url': conf['schema.registry.url'],
+        'schema.registry.basic.auth.credentials.source': conf['basic.auth.credentials.source'],
+        'schema.registry.basic.auth.user.info': conf['schema.registry.basic.auth.user.info'],
```
rnpridgeon, line 52: Are SR cloud credentials different from your standard ccloud credentials? If not it might be worth showing an example of the `SASL_INHERIT` provider which would be much more convenient.

clients/cloud/python/consumer_ccsr.py
```diff
@@ -0,0 +1,94 @@
...
+                name_object = record_key.name
+                name = name_object['name']
+                record_value = ccloud_lib.Count(msg.value())
+                count_object = record_value.count
+                count = count_object['count']
+                total_count += count
+                print("Consumed record with key {} and value {}, \
+                      and updated total count to {}"
+                      .format(name, count, total_count))
+            else:
```
rnpridgeon: This is easier to read if you instead check for a the error then continue like you do with `msg is None`; leaving the happy path logic out of the if statement.

clients/cloud/python/producer_ccsr.py
```diff
@@ -0,0 +1,105 @@
...
+           'sasl.password': conf['sasl.password'],
+           'schema.registry.url': conf['schema.registry.url'],
+           'schema.registry.basic.auth.credentials.source': conf['basic.auth.credentials.source'],
+           'schema.registry.basic.auth.user.info': conf['schema.registry.basic.auth.user.info']
+    }, default_key_schema=ccloud_lib.schema_key, default_value_schema=ccloud_lib.schema_value)
+
+    # Create topic if needed
+    # Examples of additional admin API functionality:
+    # https://github.com/confluentinc/confluent-kafka-python/blob/master/examples/adminapi.py
+    a = AdminClient({
```
rnpridgeon: If we put Admin client in ccloud_lib we could reuse it in future examples.

rnpridgeon (approved):
Few small comments otherwise LGTM

clients/cloud/python/README.md
ybyzek: Do you mean https://github.com/confluentinc/confluent-kafka-python/pull/578 ?
Reasons to NOT mention it:

1. It results in a warning and does NOT impact functionality
2. Slippery slope: this isn't the place to mention regressions/bugs/etc

clients/cloud/python/consumer_ccsr.py
ybyzek: They are different

clients/cloud/python/producer_ccsr.py
ybyzek: Excellent suggestion.  I actually already have a need now -- two different producers use the AdminClient code.

clients/cloud/python/README.md
rnpridgeon: Reasons to mention it: 
1. WARN messages scare new users in the first 15 
2. It is not the place for all bugs no but it is the place to people who are unfamiliar with our clients and possibly python. These types of errors throw new comers off and can send them  searching for an issue that may not be there for hours. This arguable makes for a terrible first 15. 

Anyway I'll leave it to your discretion.

clients/cloud/python/consumer_ccsr.py
ybyzek: Done

clients/cloud/python/README.md
ybyzek: Compromise: I added the warning message to confluentinc/confluent-kafka-python#578 so at least it's searchable.

PR #159, opened 2019-05-10
DEVX-833: Kinesis -> Confluent Cloud -> cloud storage

ybyzek:


kinesis-s3/read-data.sh
```diff
@@ -0,0 +1,32 @@
...
+export KAFKA_LOG4J_OPTS="-Dlog4j.rootLogger=DEBUG,stdout -Dlog4j.logger.kafka=DEBUG,stdout" && timeout 10 confluent consume $KAFKA_TOPIC_NAME_OUT --cloud --from-beginning --property print.key=true --value-format avro --property basic.auth.credentials.source=${BASIC_AUTH_CREDENTIALS_SOURCE} --property schema.registry.basic.auth.user.info=${SCHEMA_REGISTRY_BASIC_AUTH_USER_INFO} --property schema.registry.url=${SCHEMA_REGISTRY_URL} --property key.deserializer=org.apache.kafka.common.serialization.StringDeserializer
+
+echo -e "\nS3 data:"
+if [[ ! -f avro-tools-1.8.2.jar ]]; then
+  curl -L http://mirror.metrocast.net/apache/avro/avro-1.8.2/java/avro-tools-1.8.2.jar --output avro-tools-1.8.2.jar
+fi
+for key in $(aws s3api list-objects --bucket confluent-kafka-connect-s3-demo | jq -r '.Contents[].Key'); do
```
ghost: There's a hard-coded S3 bucket name here

kinesis-s3/stop.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+# Source library 
+. ../utils/helper.sh
+
+source config/demo.cfg
+SCHEMA_REGISTRY_CONFIG_FILE=$HOME/.ccloud/config
```
ghost: This script is defaulting to cloud SR, while `start.sh` defaults to local SR

kinesis-s3/start.sh
```diff
@@ -0,0 +1,114 @@
...
+cat <<EOF >> $CONNECT_CONFIG
+rest.port=$CONNECT_REST_PORT
+rest.advertised.name=connect-cloud
+rest.hostname=connect-cloud
+group.id=connect-cloud
+request.timeout.ms=20000
+retry.backoff.ms=500
+plugin.path=$CONFLUENT_HOME/share/java,$CONFLUENT_HOME/share/confluent-hub-components
+EOF
+export CLASSPATH=$CONFLUENT_HOME/kafka-connect-replicator/kafka-connect-replicator-5.2.1.jar && connect-distributed $CONNECT_CONFIG > $CONFLUENT_CURRENT/connect/connect-ccloud.stdout 2>&1 &
```
ghost: Why is Replicator jar being added to the classpath in this demo? Doesn't seem like it's needed here, but maybe I'm missing something

kinesis-s3/start.sh
```diff
@@ -0,0 +1,114 @@
...
+#echo "Sleeping 40 seconds waiting for Connect to start"
+#sleep 40
+
+# Setup Kinesis streams
+aws kinesis create-stream --stream-name $KINESIS_STREAM_NAME --shard-count 1
```
ghost: This will create a Kinesis stream in the default AWS region for the user, not necessarily the same region configured explicitly for other commands below with `$DEMO_REGION_LC` and `$DEMO_REGION_UC`

kinesis-s3/stop.sh
```diff
@@ -0,0 +1,41 @@
...
+check_env || exit 1
+
+aws kinesis delete-stream --stream-name $KINESIS_STREAM_NAME
+aws s3 rm --recursive s3://$DEMO_BUCKET_NAME/topics/${KAFKA_TOPIC_NAME_OUT}
+aws s3 rm --recursive s3://$DEMO_BUCKET_NAME/topics/COUNT_PER_CITY
+rm -f data.avro
+
+# Delete topics in Confluent Cloud
+topics=$(ccloud topic list)
+topics_to_delete="$KAFKA_TOPIC_NAME_IN $KAFKA_TOPIC_NAME_OUT COUNT_PER_CITY connect-configs connect-statuses connect-offsets"
```
ghost: Should `LOCATIONS` be in this list of topics to delete?

kinesis-s3/read-data.sh
ybyzek: Good catch, thanks

kinesis-s3/stop.sh
ybyzek: That was intentional as I'm toggling between the two at the moment (we can discuss why offline).

Long term: it will be CCSR

kinesis-s3/start.sh
ybyzek: @mikebin -- at the time I added it because otherwise it was generating an error, but I can't reproduce it.  The script has evolved quite a bit and now seems to work with out it.  I'll remove it since it appears to work without it now.  Thanks for mentioning this!

kinesis-s3/stop.sh
ybyzek: There is no `LOCATIONS` topic to delete.  The stream of that name has the demo topic name:

```bash
ksql> show streams;

 Stream Name | Kafka Topic       | Format    
---------------------------------------------
 LOCATIONS   | kinesis-s3-topic1 | DELIMITED 
---------------------------------------------
```

mikebin:
Sorry about that mistake. I thought I saw something while running the start
script that said `LOCATIONS` existed already, but maybe I didn’t cleanly
stop everything first. I’ll follow up if I’m able to reproduce

kinesis-s3/start.sh
ybyzek: Good catch.  I've also optimized the code by not requiring the user to specify `DEMO_REGION_UC`

kinesis-cloud/read-data.sh
```diff
@@ -0,0 +1,40 @@
...
+
+source config/demo.cfg
+DELTA_CONFIGS_DIR=delta_configs
+source $DELTA_CONFIGS_DIR/env.delta
+
+check_env || exit 1
+check_aws || exit
+
+echo -e "\n\nKinesis data:"
+while read -r line ; do echo "$line" | base64 -id; echo; done <<< "$(aws kinesis get-records --shard-iterator $(aws kinesis get-shard-iterator --shard-id shardId-000000000000 --shard-iterator-type TRIM_HORIZON --stream-name $KINESIS_STREAM_NAME --query 'ShardIterator') | jq '.Records[].Data')"
```
ghost: Might be good for this command to specify the region configured in `config/demo.cfg` in case the user's default region does not match

ghost (approved):
One minor comment added, but otherwise LGTM

kinesis-cloud/read-data.sh
ybyzek: Agree, and done!

PR #160, opened 2019-05-13
Fix typo in Streams App Tutorial docs

ableegoldman:
The "Test" suffix is currently missing from the test names in exercises 2-6 of https://docs.confluent.io/current/tutorials/examples/microservices-orders/docs/ which causes the test/build to fail when following the instructions to step through the exercises. I have verified that the tests will pass following the corrected commands in this PR


ableegoldman:
@ybyzek

PR #170, opened 2019-06-05
Confluent Cloud client example: Rust

frankgreco:
@confluentinc/devx


ybyzek:
@frankgreco can you please add Rust to https://github.com/confluentinc/examples/blob/5.2.1-post/clients/cloud/README.md ?

clients/cloud/rust/README.md
```diff
@@ -0,0 +1,52 @@
...
+# Example 1: Hello World!
+In this example, the producer writes Kafka data to a topic in Confluent Cloud. 
+Each record has a key representing a username (e.g. `alice`) and a value.
+The consumer reads the same topic from Confluent Cloud.
+
+1. Run the producer, passing in arguments for (a) the local file with configuration parameters to connect to your Confluent Cloud instance and (b) the topic name:
+
+```bash
+$ RUSTFLAGS='-C link-args=-lzstd' cargo build
+$ ./target/debug/producer --config ~/.ccloud/example.config --topic i_love_rust
```
ybyzek: For consistency with all the other client examples, let's have topic name `test1`

clients/cloud/rust/README.md
```diff
@@ -0,0 +1,52 @@
+# Overview
+
+Produce messages to produce and consume messages from [Confluent Cloud](https://www.confluent.io/confluent-cloud/) using the [rust-rdkafka client for Apache Kafka](https://github.com/fede1024/rust-rdkafka).
+
+# Prerequisites
+* [Rust Client for Apache Kafka](https://github.com/fede1024/rust-rdkafka#installation) installed on your machine
```
ybyzek: `Client` -> `client`

clients/cloud/rust/src/producer.rs
```diff
@@ -0,0 +1,44 @@
...
+use rdkafka::message::OwnedMessage;
+use rdkafka::producer::{FutureProducer, FutureRecord};
+use std::boxed::Box;
+
+mod utils;
+
+fn log_produce_result(result: Result<(i32, i64), (KafkaError, OwnedMessage)>) -> Result<(), ()> {
+    result
+        .and_then(|(p, o)| {
+            println!("produced to partition {} with offset {}", p, o);
```
ybyzek: For consistency with the other client examples, can this print topic name as well?  Specifically, this is what the other clients print:

```bash
Produced record to topic test1 partition [0] @ offset 0
...
```

clients/cloud/rust/src/producer.rs
```diff
@@ -0,0 +1,44 @@
...
+        .map_err(|(err, _)| eprintln!("kafka error: {}", err))
+}
+
+fn main() -> Result<(), Box<std::error::Error>> {
+    let (topic, config) = utils::get_config()?;
+    let producer: FutureProducer = config.create()?;
+
+    let messages = (0..9)
+        .map(|msg| {
+            producer.send(
```
ybyzek, line 47: For consistency with the other client examples, can this print that it's _about to send_?  Specifically, this is what the other clients print:

```bash
Producing record: alice 	 {"count": 0}
...
```

clients/cloud/rust/src/consumer.rs
```diff
@@ -0,0 +1,57 @@
...
+use rdkafka::Message;
+use std::boxed::Box;
+use tokio::runtime::current_thread::Runtime;
+
+mod utils;
+
+fn echo_message<M: Message>(msg: M) -> Result<(), std::str::Utf8Error> {
+    let deserialize = |o| match o {
+        None => Ok("".to_string()),
+        Some(val) => Ok(std::str::from_utf8(val)?.to_string()),
```
gardnervickers: I think this can just be `String::from_utf8(val)`

clients/cloud/rust/src/consumer.rs
```diff
@@ -0,0 +1,57 @@
...
+    let processor = consumer
+        .start()
+        .filter_map(|result| match result {
+            Ok(_) => result.ok(),
+            Err(err) => {
+                eprintln!("error consuming from message stream: {}", err);
+                None
+            }
+        })
+        .for_each(move |msg| {
```
gardnervickers: Hmm do we need the `move` here?

clients/cloud/README.md
```diff
@@ -22,7 +22,7 @@ The following subset includes examples with Confluent Cloud Schema Registry and
 | [![](images/kafka-connect-datagen.png)](kafka-connect-datagen/) | [ksql-datagen](ksql-datagen/) | |
 
 
-# Other Confluent Cloud Demos
+# Other Confluent Cloud Dem
```
ybyzek: @frankgreco -- typo

frankgreco: fixed

clients/cloud/README.md
```diff
@@ -10,7 +10,7 @@ This directory includes examples of Kafka client applications connecting to [Con
 | [![](images/scala.png)](scala/)   | [![](images/confluent-cli.png)](confluent-cli/) | [![](images/ruby.png)](ruby/)     |
 | [![](images/groovy.png)](groovy/) | [![](images/kotlin.png)](kotlin/)               | [![](images/nodejs.png)](nodejs/) |
 | [![](images/kafkacat.jpg)](kafkacat/) | [![](images/dotnet.png)](csharp/)           | [![](images/c.png)](c/) |
-| [![](images/kafka-connect-datagen.png)](kafka-connect-datagen/) | [ksql-datagen](ksql-datagen/) | |
+| [![](images/kafka-connect-datagen.png)](kafka-connect-datagen/) | [ksql-datagen](ksql-datagen/) | [<img src="images/rust.png" width="100" />](rust/) |
```
edenhill: This image import should probably not be different from the existing ones.

clients/cloud/rust/.gitignore
```diff
@@ -0,0 +1,2 @@
+/target
```
edenhill, line 1: This should be just target, right? /target is outside the git repo

clients/cloud/rust/README.md
```diff
@@ -0,0 +1,61 @@
+# Overview
+
+Produce messages to produce and consume messages from [Confluent Cloud](https://www.confluent.io/confluent-cloud/) using the [rust-rdkafka client for Apache Kafka](https://github.com/fede1024/rust-rdkafka).
```
edenhill: "Produce messages to produce .." is redundant, should probably be "Produce messages to and consume messages from .."

clients/cloud/rust/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+```bash
+$ cat ~/.ccloud/example.config
+bootstrap.servers=<broker-1,broker-2,broker-3>
+sasl.username=<api-key-id>
+sasl.password=<secret-access-key>
+```
+
+# Example 1: Hello World!
+In this example, the producer writes Kafka data to a topic in Confluent Cloud.
```
edenhill: "writes data to a Kafka topic"

clients/cloud/rust/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+# Example 1: Hello World!
+In this example, the producer writes Kafka data to a topic in Confluent Cloud. 
+Each record has a key representing a username (e.g. `alice`) and a value.
+The consumer reads the same topic from Confluent Cloud.
+
+1. Run the producer, passing in arguments for (a) the local file with configuration parameters to connect to your Confluent Cloud instance and (b) the topic name:
+
+```bash
+$ RUSTFLAGS='-C link-args=-lzstd' cargo build
```
edenhill: Why are these RUSTFLAGS needed? librdkafka should already be linked with -lzstd

clients/cloud/rust/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "confluent_cloud"
```
edenhill: confluent_cloud_example to avoid confusion with future confluent_cloud packages (whatever that would be)?

clients/cloud/rust/README.md
```diff
@@ -0,0 +1,61 @@
...
+Successfully produced record to topic test1 partition [5] @ offset 23
+Successfully produced record to topic test1 partition [5] @ offset 24
+Successfully produced record to topic test1 partition [5] @ offset 25
+Successfully produced record to topic test1 partition [5] @ offset 26
+```
+
+2. Run the consumer, passing in arguments for (a) the local file with configuration parameters to connect to your Confluent Cloud instance and (b) the same topic name as used above. Verify that the consumer received all the messages:
+
+```bash
+$ RUSTFLAGS='-C link-args=-lzstd' cargo build
```
edenhill: This repeats the build step. Maybe put the build as a separate step before running the producer and cosnumer?

clients/cloud/rust/README.md
```diff
@@ -0,0 +1,61 @@
...
+$ ./target/debug/consumer --config ~/.ccloud/example.conf --topic test1
+Consumed record from topic test1 partition [5] @ offset 99 with key alice and value 0
```
edenhill: nit: these offsets don't match the producer offsets above.

clients/cloud/rust/src/consumer.rs
```diff
@@ -0,0 +1,57 @@
+extern crate futures;
```
edenhill: add copyright blob

clients/cloud/rust/src/consumer.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    consumer.subscribe(&vec![topic.as_ref()])?;
+
+    let processor = consumer
+        .start()
+        .filter_map(|result| match result {
+            Ok(_) => result.ok(),
+            Err(err) => {
+                eprintln!("error consuming from message stream: {}", err);
+                None
```
edenhill, line 55: is this aborting the consumption or will it keep going? it should be doing the latter

clients/cloud/rust/src/consumer.rs
```diff
@@ -0,0 +1,57 @@
...
+            }
+        })
+        .for_each(move |msg| {
+            echo_message(msg).map_err(|_| eprintln!("error unserializating message"))
```
edenhill: deserializing :)

clients/cloud/rust/src/producer.rs
```diff
@@ -0,0 +1,52 @@
+extern crate futures;
```
edenhill: copyright blobbage

clients/cloud/rust/src/producer.rs
```diff
@@ -0,0 +1,52 @@
...
+    let (topic, config) = utils::get_config()?;
+    let producer: FutureProducer = config.create()?;
+
+    let messages = (0..9)
+        .map(|msg| {
+            let value = msg.to_string();
+            println!("Preparing to produce record: {} {}", "alice", value);
+            producer.send(
+                FutureRecord::to(&topic)
+                    .payload(value.as_bytes())
```
edenhill, line 49: Should preferably be a conversion to utf-8 here, just to be clear that serialization is indeed a good idea.

clients/cloud/rust/src/producer.rs
```diff
@@ -0,0 +1,52 @@
...
+                    .key("alice"),
```
edenhill, line 50: dito

clients/cloud/rust/src/utils.rs
```diff
@@ -0,0 +1,53 @@
+extern crate clap;
```
edenhill: copyright blobbing

clients/cloud/rust/src/utils.rs
```diff
@@ -0,0 +1,53 @@
...
+            Arg::with_name("topic")
+                .help("test topic to use")
+                .long("topic")
+                .takes_value(true)
+                .required(true),
+        )
+        .get_matches();
+
+    let mut kafka_config = ClientConfig::new();
+    kafka_config.set_log_level(RDKafkaLogLevel::Debug);
```
edenhill: Not needed. No debug is logged unless you specify a "debug" config property, in which case it also raises the log level.

clients/cloud/rust/src/utils.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    let file = File::open(matches.value_of("config").ok_or("error parsing config")?)?;
+    for line in BufReader::new(&file).lines() {
+        let cur_line = line?;
+        if cur_line.starts_with('#') || cur_line.chars().count() < 1 {
```
edenhill: Might want to l/r trim the line.

clients/cloud/rust/src/utils.rs
```diff
@@ -0,0 +1,53 @@
...
+            continue;
+        }
+        let key_value: Vec<&str> = cur_line.split("=").collect();
+        kafka_config.set(
```
edenhill, line 51: will this propagate errors?

clients/cloud/rust/src/consumer.rs
```diff
@@ -0,0 +1,57 @@
...
+        deserialize(msg.key())?,
+        deserialize(msg.payload())?,
+    );
+
+    Ok(())
+}
+
+fn main() -> Result<(), Box<std::error::Error>> {
+    let (topic, mut config) = utils::get_config()?;
+    let consumer: StreamConsumer = config.set("group.id", "rust_example_group_1").create()?;
```
edenhill, line 45: it needs some more config for cloud, such as "security.protocol" and "sasl.mechanisms". Same for the producer

edenhill (changes requested):

clients/cloud/README.md
frankgreco: the reason it is is because this is the only way in GH markdown that I know of to resize an image.

clients/cloud/rust/.gitignore
frankgreco: this `.gitignore` was generated by the cargo cli

clients/cloud/rust/README.md
frankgreco: copy/paste from other client examples - I left it as is for uniformity

clients/cloud/rust/README.md
frankgreco: copy/paste from other client examples - I left it as is for uniformity

I guess I could change in all client example readme's

clients/cloud/rust/README.md
frankgreco: I would love to figure out how to do a build without specifying this but I spend too much time trying to get it to work and this was the workaround. Here is the error I get:

```
  = note: Undefined symbols for architecture x86_64:
            "_ZSTD_createCStream", referenced from:
                _rd_kafka_zstd_compress in librdkafka_sys-347d93b86e02028e.rlib(rdkafka_zstd.o)
            "_ZSTD_initCStream", referenced from:
                _rd_kafka_zstd_compress in librdkafka_sys-347d93b86e02028e.rlib(rdkafka_zstd.o)
            "_ZSTD_endStream", referenced from:
                _rd_kafka_zstd_compress in librdkafka_sys-347d93b86e02028e.rlib(rdkafka_zstd.o)
            "_ZSTD_decompress", referenced from:
                _rd_kafka_zstd_decompress in librdkafka_sys-347d93b86e02028e.rlib(rdkafka_zstd.o)
            "_ZSTD_compressBound", referenced from:
                _rd_kafka_zstd_compress in librdkafka_sys-347d93b86e02028e.rlib(rdkafka_zstd.o)
            "_ZSTD_isError", referenced from:
                _rd_kafka_zstd_decompress in librdkafka_sys-347d93b86e02028e.rlib(rdkafka_zstd.o)
                _rd_kafka_zstd_compress in librdkafka_sys-347d93b86e02028e.rlib(rdkafka_zstd.o)
            "_ZSTD_compressStream", referenced from:
                _rd_kafka_zstd_compress in librdkafka_sys-347d93b86e02028e.rlib(rdkafka_zstd.o)
            "_ZSTD_freeCStream", referenced from:
                _rd_kafka_zstd_compress in librdkafka_sys-347d93b86e02028e.rlib(rdkafka_zstd.o)
            "_ZSTD_getFrameContentSize", referenced from:
                _rd_kafka_zstd_decompress in librdkafka_sys-347d93b86e02028e.rlib(rdkafka_zstd.o)
            "_ZSTD_getErrorCode", referenced from:
                _rd_kafka_zstd_decompress in librdkafka_sys-347d93b86e02028e.rlib(rdkafka_zstd.o)
            "_ZSTD_getErrorName", referenced from:
                _rd_kafka_zstd_decompress in librdkafka_sys-347d93b86e02028e.rlib(rdkafka_zstd.o)
                _rd_kafka_zstd_compress in librdkafka_sys-347d93b86e02028e.rlib(rdkafka_zstd.o)
          ld: symbol(s) not found for architecture x86_64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

clients/cloud/rust/README.md
frankgreco: see above comment

clients/cloud/rust/README.md
frankgreco: haha good point :)

clients/cloud/rust/src/consumer.rs
frankgreco: do you mean a confluent one that we have in the other examples?

clients/cloud/rust/src/consumer.rs
frankgreco: those are provided by the passed in config

clients/cloud/rust/src/utils.rs
frankgreco: anywhere where there is an error, a result is returned and errors are propagated

clients/cloud/rust/src/consumer.rs
frankgreco: it will keep going - it's just a filter

clients/cloud/rust/src/producer.rs
frankgreco: in rust, all Strings are utf-8 encoded: https://doc.rust-lang.org/std/string/struct.String.html

clients/cloud/README.md
edenhill: Okay, can we resize the actual image instead? We did this with the other ones

clients/cloud/rust/README.md
edenhill: Not sure that applies here though, this just seems to rebuild the same thing again?

clients/cloud/rust/src/consumer.rs
edenhill: yeah

clients/cloud/rust/src/consumer.rs
edenhill: But the config file example above does not have them.

clients/cloud/README.md
frankgreco: yea sure I could do that - 256x256 seems like it should fit

clients/cloud/rust/README.md
frankgreco: oh I see what you mean - yea you're right this step is redundant

clients/cloud/rust/src/consumer.rs
frankgreco: good point - i'll make it clear in the readme that they should be there by adding them to the example

frankgreco: done

clients/cloud/rust/src/utils.rs
frankgreco: done

clients/cloud/rust/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "rust_kafka_client_example"
```
gardnervickers: I think the convention is to use dashes instead of underscores here

clients/cloud/rust/README.md
gardnervickers: This looks like a bug, it should be possible to automate this with a `build.rs` file
https://github.com/fede1024/rust-rdkafka/issues/125

clients/cloud/rust/src/consumer.rs
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+extern crate futures;
+extern crate rdkafka;
+extern crate tokio;
```
gardnervickers: It might be worth considering using Rust 2018 and removing these `extern crate` statements.

clients/cloud/rust/src/consumer.rs
```diff
@@ -0,0 +1,72 @@
...
+        msg.partition(),
+        msg.offset(),
+        deserialize(msg.key())?,
+        deserialize(msg.payload())?,
+    );
+
+    Ok(())
+}
+
+fn main() -> Result<(), Box<std::error::Error>> {
```
gardnervickers: Should this be `dyn std::error::Error`?

clients/cloud/rust/src/producer.rs
```diff
@@ -0,0 +1,67 @@
...
+use futures::future::Future;
+use rdkafka::error::KafkaError;
+use rdkafka::message::OwnedMessage;
+use rdkafka::producer::{FutureProducer, FutureRecord};
+use std::boxed::Box;
+
+mod utils;
+
+fn log_produce_result(
+    topic: String,
```
gardnervickers: Does topic here need to be owned or can it be `&str`?

clients/cloud/rust/src/producer.rs
```diff
@@ -0,0 +1,67 @@
...
+                    .key("alice"),
+                0,
+            )
+        })
+        .collect::<Vec<_>>();
+
+    for msg in messages {
+        msg.wait()
+            .map_err(|err| eprintln!("error producing message: {}", err))
+            .and_then(|result| log_produce_result(topic.to_owned(), result))
```
gardnervickers: Im not sure `topic` needs to be cloned here if `log_produce_result` does not need an owned String.

clients/cloud/rust/README.md
frankgreco: @gardnervickers good find! let me try that 💯

clients/cloud/rust/src/consumer.rs
frankgreco: I got the inspiration from [here](https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/question-mark-in-main-and-tests.html).

clients/cloud/rust/src/consumer.rs
frankgreco: Nice! Didn't know these weren't required in Rust 2018 :)

clients/cloud/rust/src/producer.rs
frankgreco: You are right, it doesn't need to be owned!

clients/cloud/rust/src/consumer.rs
frankgreco: If I do that, then I need to use `.to_vec()` which then means I need add type annotation to the closure. Hence, it's less code to have it as is.

clients/cloud/rust/src/consumer.rs
frankgreco: nope :) good catch - I wish the compiler told you this because at one point I probably did but then when I no longer needed it it's still valid code and doesn't warn.

clients/cloud/rust/src/consumer.rs
gardnervickers: Hmm on second thought, we're taking a `&[u8]` here, converting it to a `&str` and then to a `String`. I think we should be able to avoid the `to_string()` copying the `&str` to a `String` here right?

frankgreco: ah good observation - no need to convert to a `String`

frankgreco:
@edenhill @ybyzek @gardnervickers I believe I have addressed all your suggestions 😄

ybyzek (approved):
@frankgreco thanks for the PR!  Can you resize the Rust graphic just a tough smaller (shave off ~15%), but otherwise LGTM!  I approve.

clients/cloud/rust/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name    = "rust-kafka-client-example"
+version = "0.1.0"
+authors = ["GRECO, FRANK <frank@confluent.io>"]
+edition = "2018"
```
edenhill, line 5: 2019

edenhill (approved):
LGTM!

clients/cloud/rust/Cargo.toml
frankgreco: that's not the current year, but the rust edition

clients/cloud/rust/src/producer.rs
```diff
@@ -0,0 +1,64 @@
...
+use rdkafka::message::OwnedMessage;
+use rdkafka::producer::{FutureProducer, FutureRecord};
+use std::boxed::Box;
+
+mod utils;
+
+fn log_produce_result(
+    topic: &str,
+    result: Result<(i32, i64), (KafkaError, OwnedMessage)>,
+) -> Result<(), ()> {
```
gardnervickers, line 27: If there is nothing to return for the result, do we even need to return anything?

clients/cloud/rust/src/producer.rs
```diff
@@ -0,0 +1,64 @@
...
+                0,
+            )
+        })
+        .collect::<Vec<_>>();
+
+    for msg in messages {
+        msg.wait()
+            .map_err(|err| eprintln!("error producing message: {}", err))
+            .and_then(|result| log_produce_result(&topic, result))
+            .ok();
```
gardnervickers, line 60: Is this `.ok()` necessary?

clients/cloud/rust/src/consumer.rs
```diff
@@ -0,0 +1,66 @@
...
+        None => Ok(""),
+        Some(val) => Ok(std::str::from_utf8(val)?),
+    };
+
+    println!(
+        "Consumed record from topic {} partition [{}] @ offset {} with key {} and value {}",
+        msg.topic(),
+        msg.partition(),
+        msg.offset(),
+        deserialize(msg.key())?,
```
gardnervickers, line 36: This could just be
```
    msg.key()
        .map(|s| std::str::from_utf8(s)?)
        .unwrap_or("");
```

gardnervickers (commented):
LGTM, one thing to watch out for is that trait objects need a `dyn` before them, bare trait objects are deprecated. 
So `Box<std::error::Error>` should become `Box<dyn std::error::Error>`

clients/cloud/rust/src/consumer.rs
frankgreco: yea but then I have to have that same code for the value which is why I abstracted it out into a function.

clients/cloud/rust/src/producer.rs
frankgreco: I need to if I want to use it in an `and_then` like I do

clients/cloud/rust/src/producer.rs
frankgreco: yes because i'm using `and_then` which returns a Result which means that there could be an error. We know there won't be because the result we're returning doesn't throw any error but we still want to make the compiler happy

gardnervickers: I'm not sure I follow, essentially we're going from ` Result<Result<(i32, i64), (KafkaError, OwnedMessage)>>` here to `Option<()>` but not consuming the option. That should be equivalent to `Result<Result<(i32, i64), (KafkaError, OwnedMessage)>> -> ()` (since we don't use the option. Calling `expect()` here would be fine since it's illegal for `log_produce_result` to return any error.

frankgreco: i'm using `Result::and_then` not `Option::and_then` so a `Result` will be returned, not an `Option`

gardnervickers: Hmm, the `.ok()` here will consume the result and return an option.

frankgreco: yea basically I'm saying, "even if the `and_then` has an error in the result (which it won't), it's okay anyways". That way, the compile won't complain about an unchecked result.

gardnervickers (approved):

PR #172, opened 2019-06-06
Test for curl being installed

### LINKED ISSUE #158: check_env doesn't validate curl is installed
check_env should validate curl is installed. A minimal Ubuntu 18.04.2 LTS installation does not install curl by default.

```
 $ check_running_cp 5.2

This script expects Confluent Platform version 5.2 but the running version is .
To proceed please either: change the examples repo branch to  or update the running Confluent Platform to version 5.2.
```

If you look at the output of `confluent` you can see the error ...

```
$ confluent version
'confluent' requires 'curl'.
````

### PR DESCRIPTION
**Pull request by ghost**
resolves #158

### DIFF
--- a/utils/helper.sh
+++ b/utils/helper.sh
@@ -1,6 +1,11 @@
 #!/bin/bash
 
 function check_env() {
+  if [[ $(type curl 2>&1) =~ "not found" ]]; then
+    echo "'curl' is not found.  Install curl to continue"
+    exit 1
+  fi
+
   if [[ -z "$CONFLUENT_HOME" ]]; then
     echo "\$CONFLUENT_HOME is not defined. Run 'export CONFLUENT_HOME=/path/to/confluentplatform' and try again"
     exit 1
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: utils/helper.sh [hunk 1]
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash
 
 function check_env() {
+  if [[ $(type curl 2>&1) =~ "not found" ]]; then
```
ybyzek: 1. Let's pull the curl check out into its own function `check_curl()`.
2. Since the curl is called from `confluent version`, then `check_curl()` should be called from `check_running_cp()`

ghost: Done.

There's other functions in that file that also run curl, so we could also use this function there - eg. `check_running_elasticsearch()` and `check_running_grafana()`.  Not sure if they ever get called without having first called `check_running_cp()` though.

ybyzek (approved):
Thanks for the PR @WheresAlice !

Issue #175, opened 2019-06-13
demo-consumer-1 group is hard coded in java consumer client so fails on ACL


ScottGriz:
https://github.com/confluentinc/examples/blob/5.2.1-post/ccloud/acl.sh
Line 315
ccloud kafka acl create --allow --service-account-id $SERVICE_ACCOUNT_ID --operation READ --consumer-group java_example_group_1
Creates an ACL for java_example_group_1 but the java client hard codes group to demo-consumer-1 

Caused by: org.apache.kafka.common.errors.GroupAuthorizationException: Not authorized to access group: demo-consumer-1
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 7.750 s
[INFO] Finished at: 2019-06-13T18:27:42+00:00
[INFO] Final Memory: 26M/377M
https://github.com/confluentinc/examples/blob/5.2.1-post/ccloud/acl.sh 
https://github.com/confluentinc/examples/blob/5.2.1-post/ccloud/acl.sh
https://github.com/confluentinc/examples/blob/5.2.1-post/ccloud/acl.sh
Line 315 ccloud kafka acl create --allow --service-account-id $SERVICE_ACCOUNT_ID --

https://github.com/confluentinc/examples/blob/5.2.1-post/clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/ConsumerExample.java
I changed the group.id to match the ACL
    props.put(ConsumerConfig.GROUP_ID_CONFIG, "demo-consumer-1"); CHANGED to 
 but we never saw the expected output "Successfully joined group" so acl.sh threw an error

[ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:1.2.1:java (default-cli) on project clients-example: An exception occured while executing the Java class. null: InvocationTargetException: Not authorized to access group: demo-consumer-1 -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch

ybyzek:
Resolved in #176

Issue #174, opened 2019-06-13
Cloud ACL script has a bug in parsing the API Secret 


ScottGriz:
https://github.com/confluentinc/examples/blob/5.2.1-post/ccloud/acl.sh . 

Line 164 API_SECRET_SA=$(echo "$OUTPUT" | grep "\| Secret" | awk '{print $4;}')
The Double quotes around $OUTPUT cause a bogus string to be put in the client.config
I changed the code to 
API_SECRET_SA=$(echo "$OUTPUT" | grep '| Secret' | awk '{print $4;}')
Same as the API_KEY and now  works.  client.config is built with the correct API_SECRET_SA value

ybyzek:
Resolved in #176

PR #176, opened 2019-06-14
DEVX-891: address issues #174 and #175

ybyzek:


ccloud/acl.sh
```diff
@@ -121,6 +121,14 @@ if [[ $? != 0 ]]; then
   exit 1
 fi
 
+echo -e "\n# Verify cluster is a Confluent Cloud Enterprise cluster"
+echo "ccloud kafka acl list"
+OUTPUT=$(ccloud kafka acl list 2>&1)
```
ybyzek, line 126: @codyaray @DABH : is there a better way to check if a Kafka cluster is CCE or CCP?

rmoff (approved):

PR #191, opened 2019-07-03
DEVX-868: RBAC: kafka broker

ybyzek:


rnpridgeon (approved):
LGTM although it looks to be a work in progress still based on some of the non diff lines.

Have you considered using Exit Traps to initialize clean up? Might be something worth looking into nut certainly not a must do.

ybyzek:
@rnpridgeon -- thank you.  Exit traps are a fantastic idea.

In order to unblock and move forward with the demo, I will merge this PR and I have filed a separate jira to track adding exit traps.

PR #195, opened 2019-07-09
DEVX-915: RBAC demo v1: Connect

ybyzek:


security/rbac/delta_configs/connect-avro-distributed.properties.delta
```diff
@@ -1,37 +1,19 @@
...
 sasl.login.callback.handler.class=io.confluent.kafka.clients.plugins.auth.token.TokenUserLoginCallbackHandler
-
 sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required username="connect" password="connect1" metadataServerUrls="http://localhost:8090";
 
 ## Connector client (producer, consumer, admin client) properties ##
 
 # Allow producer/consumer/admin client overrides (this enables per-connector principals)
 connector.client.config.override.policy=All
 
-# Default producer configs to use for connectors
```
C0urante, line 16: I think we can (and should) retain these. The only property that we _don't_ want to set in the worker file is `sasl.jaas.config`, which provides the actual credentials that identify the service account. Since we want each connector to have its own service account, there shouldn't be a default for that.

However, everything else (`security.protocol`, `sasl.mechanism`, and `sasl.login.callback.handler`) just specifies how authentication will happen (but not user credentials), so it's fine to set those as global defaults that every connector will inherit unless explicitly overridden.

security/rbac/delta_configs/connect-avro-distributed.properties.delta
```diff
@@ -53,10 +35,13 @@ confluent.metadata.http.auth.credentials.provider=BASIC
...
+config.providers.secret.param.kafkastore.topic=_secrets
 config.providers.secret.param.kafkastore.bootstrap.servers=localhost:9092
-# Can also be SASL_SSL
-config.providers.secret.param.kafkastore.security.protocol=SASL_PLAINTEXT config.providers.secret.param.kafkastore.sasl.mechanism=OAUTHBEARER
+config.providers.secret.param.kafkastore.security.protocol=SASL_PLAINTEXT
+config.providers.secret.param.kafkastore.sasl.mechanism=OAUTHBEARER
 config.providers.secret.param.kafkastore.sasl.login.callback.handler.class=io.confluent.kafka.clients.plugins.auth.token.TokenUserLoginCallbackHandler
 config.providers.secret.param.kafkastore.sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required username="connect" password="connect1" metadataServerUrls="http://localhost:8090";
+
+rest.advertised.host.name=localhost
```
C0urante: OOC, do things break if this isn't specified?

security/rbac/delta_configs/connector-source.properties.delta
```diff
@@ -0,0 +1,3 @@
+producer.override.sasl.mechanism=OAUTHBEARER
```
C0urante: Following up on comments left on the `connect-avro-distributed.properties.delta` file--we should remove the `producer.override.sasl.mechanism` and `producer.override.sasl.login.callback.handler` configs from this file and they should instead be set in the worker file.

security/rbac/scripts/submit_datagen_pageviews_config_avro.sh
```diff
@@ -0,0 +1,34 @@
...
+    "connector.class": "io.confluent.kafka.connect.datagen.DatagenConnector",
+    "kafka.topic": "pageviews",
+    "quickstart": "pageviews",
+    "key.converter": "org.apache.kafka.connect.storage.StringConverter",
+    "value.converter": "io.confluent.connect.avro.AvroConverter",
+    "value.converter.schema.registry.url": "http://localhost:8081",
+    "value.converter.basic.auth.credentials.source": "USER_INFO",
+    "value.converter.basic.auth.user.info": "client:client1",
+    "max.interval": 100,
+    "producer.override.sasl.mechanism": "OAUTHBEARER",
```
C0urante: Following up on comments left on the `connect-avro-distributed.properties.delta` file--we should remove the `producer.override.sasl.mechanism` and `producer.override.sasl.login.callback.handler` configs from this config and they should instead be set in the worker file.

security/rbac/scripts/submit_datagen_pageviews_config_avro.sh
```diff
@@ -0,0 +1,34 @@
...
+  "name": "datagen-pageviews",
+  "config": {
+    "connector.class": "io.confluent.kafka.connect.datagen.DatagenConnector",
+    "kafka.topic": "pageviews",
+    "quickstart": "pageviews",
+    "key.converter": "org.apache.kafka.connect.storage.StringConverter",
+    "value.converter": "io.confluent.connect.avro.AvroConverter",
+    "value.converter.schema.registry.url": "http://localhost:8081",
+    "value.converter.basic.auth.credentials.source": "USER_INFO",
+    "value.converter.basic.auth.user.info": "client:client1",
```
C0urante, line 21: These credentials should match the ones in the `producer.override.sasl.jaas.config`, so that there's one service account to track for the connector that deals with both Kafka interactions (writing to, reading from, and creating topics) and Schema Registry interactions (writing to and reading from subjects).

security/rbac/scripts/enable-rbac-connect.sh
```diff
@@ -0,0 +1,152 @@
...
+# - Grant the principal User:$ADMIN_CONNECT to the ResourceOwner role for Group:connect-cluster
+# - Grant the principal User:$ADMIN_CONNECT to the ResourceOwner role for Topic:_secrets (for Secret Registry)
+# - Grant the principal User:$ADMIN_CONNECT to the ResourceOwner role for Group:secret-registry (for Secret Registry)
+##################################################
+
+# Get the Kafka cluster id
+get_cluster_id_kafka
+
+for CONNECT_TOPIC in connect-configs connect-offsets connect-statuses; do
+  echo -e "\n# Grant the principal User:$ADMIN_CONNECT to the ResourceOwner role for Topic:${CONNECT_TOPIC}"
```
C0urante: Nit: I think the language here can be tweaked. My understanding is that a [principal](https://docs.oracle.com/javase/8/docs/api/java/security/Principal.html) (at least in the Java ecosystem) is an identity like a user or a service account, so I think it might make more sense to word this like "Grant the principal <principal> the role <role> for the resource <resource>" or "Grant the principal <principal> the role <role> for the cluster <cluster>".

security/rbac/scripts/enable-rbac-connect.sh
```diff
@@ -0,0 +1,152 @@
...
+confluent-hub install --no-prompt confluentinc/kafka-connect-datagen:latest
+
+echo -e "\n# Bring up Connect"
+confluent local start connect
+
+
+##################################################
+# Connect client functions
+# - Grant the principal User:$ADMIN_CONNECT to the ResourceOwner role for Connector:$CONNECTOR_NAME
+# - Grant the principal User:connector to the DeveloperManage role for Connector:$CONNECTOR_NAME
```
C0urante: This isn't necessary; connectors don't need permissions to interact with themselves.

security/rbac/scripts/enable-rbac-connect.sh
```diff
@@ -0,0 +1,152 @@
...
+# - Grant the principal User:connector to the DeveloperWrite role for Connector:$CONNECTOR_NAME
```
C0urante: This isn't necessary; connectors don't need permissions to interact with themselves.

security/rbac/scripts/enable-rbac-connect.sh
```diff
@@ -0,0 +1,152 @@
...
+echo "confluent-hub install --no-prompt confluentinc/kafka-connect-datagen:latest"
+confluent-hub install --no-prompt confluentinc/kafka-connect-datagen:latest
+
+echo -e "\n# Bring up Connect"
+confluent local start connect
+
+
+##################################################
+# Connect client functions
+# - Grant the principal User:$ADMIN_CONNECT to the ResourceOwner role for Connector:$CONNECTOR_NAME
```
C0urante: This makes the service account for the Connect worker the ResourceOwner for the connector, which is a little strange.

Being the ResourceOwner for a connector allows someone to view its configuration, pause/resume it, create it, check its status, things like that. So that role should belong to a user, specifically, the one that issues the REST request to create the connector. Is there a user principal defined in this example?

security/rbac/scripts/enable-rbac-connect.sh
```diff
@@ -0,0 +1,152 @@
...
+echo -e "\n# List the role bindings for the principal User:connector for the Kafka cluster"
+echo "confluent iam rolebinding list --principal User:connector --kafka-cluster-id $KAFKA_CLUSTER_ID"
+confluent iam rolebinding list --principal User:connector --kafka-cluster-id $KAFKA_CLUSTER_ID
+
+echo -e "\n# List the role bindings for the principal User:connector for the Connect cluster"
+echo "confluent iam rolebinding list --principal User:connector --kafka-cluster-id $KAFKA_CLUSTER_ID --connect-cluster-id $CONNECT_CLUSTER_ID"
+confluent iam rolebinding list --principal User:connector --kafka-cluster-id $KAFKA_CLUSTER_ID --connect-cluster-id $CONNECT_CLUSTER_ID
+
+get_cluster_id_schema_registry
+echo -e "\n# Grant the principal User:$CLIENT to the ResourceOwner role for Subject:${DATA_TOPIC}-value"
```
C0urante: Should be `User:connector` instead of `User:$CLIENT` (left an in-depth explanation as a comment on the `submit_datagen_pageviews_config_avro.sh` file).

security/rbac/scripts/submit_datagen_pageviews_config.sh
```diff
@@ -0,0 +1,31 @@
...
+{
+  "name": "datagen-pageviews",
+  "config": {
+    "connector.class": "io.confluent.kafka.connect.datagen.DatagenConnector",
+    "kafka.topic": "pageviews",
+    "quickstart": "pageviews",
+    "key.converter": "org.apache.kafka.connect.storage.StringConverter",
+    "value.converter": "org.apache.kafka.connect.storage.StringConverter",
+    "max.interval": 100,
+    "producer.override.sasl.mechanism": "OAUTHBEARER",
```
C0urante: Following up on comments left on the connect-avro-distributed.properties.delta file--we should remove the producer.override.sasl.mechanism and producer.override.sasl.login.callback.handler configs from this config and they should instead be set in the worker file.

C0urante (commented):
Looks pretty good! I think there's still a bit of a disconnect on two fronts (client security configs in worker file vs. client security configs in connector file, and which principals are in play and where they should be used between the user, connectors, and the Connect worker), but the overall content is great.

tobydrake7 (commented):
Yeva, this looks great. I hope to provide review comments when I go through the steps myself later today.

security/rbac/delta_configs/connect-avro-distributed.properties.delta
ybyzek: @C0urante thanks for the explanation and it makes sense.  I've modified the code, please let me know what you think.

security/rbac/delta_configs/connect-avro-distributed.properties.delta
ybyzek: Nope!  I'll cut it.

security/rbac/delta_configs/connector-source.properties.delta
ybyzek: Agree

security/rbac/scripts/submit_datagen_pageviews_config_avro.sh
ybyzek: Agree

security/rbac/scripts/submit_datagen_pageviews_config_avro.sh
ybyzek: One SA for both the connector and SR makes logical sense, but is that the way it will always work?  It may be different, right?, depending on how SA creds are handed out?

security/rbac/delta_configs/connect-avro-distributed.properties.delta
C0urante: Looks good!

security/rbac/scripts/submit_datagen_pageviews_config_avro.sh
C0urante: While it's certainly possible, I don't think it's likely, and it goes against the recommended flow of one SA per connector. I think we should demonstrate the common case here, but I'll leave that up to you.

security/rbac/scripts/enable-rbac-connect.sh
ybyzek: Validated and removed

security/rbac/scripts/enable-rbac-connect.sh
ybyzek: Validated and removed

security/rbac/scripts/submit_datagen_pageviews_config_avro.sh
ybyzek: I'm changing everything from `connector` to `client`.

security/rbac/scripts/enable-rbac-connect.sh
ybyzek: I'm changing everything from `connector` to `client`.

security/rbac/scripts/enable-rbac-connect.sh
ybyzek: @C0urante thanks for bringing this up.  I'll pursue this separately, and it will apply to all components, not just connect

ybyzek: Resolved

security/rbac/scripts/enable-rbac-connect.sh
```diff
@@ -0,0 +1,151 @@
...
+# Get the Kafka cluster id
+get_cluster_id_kafka
+
+for CONNECT_TOPIC in connect-configs connect-offsets connect-statuses; do
+  echo -e "\n# Grant the principal User:$ADMIN_CONNECT to the ResourceOwner role for Topic:${CONNECT_TOPIC}"
+  echo "confluent iam rolebinding create --principal User:$ADMIN_CONNECT --role ResourceOwner --resource Topic:${CONNECT_TOPIC} --kafka-cluster-id $KAFKA_CLUSTER_ID"
+  confluent iam rolebinding create --principal User:$ADMIN_CONNECT --role ResourceOwner --resource Topic:${CONNECT_TOPIC} --kafka-cluster-id $KAFKA_CLUSTER_ID
+done
+
+echo -e "\n# Grant the principal User:$ADMIN_CONNECT to the ResourceOwner role for Group:connect-cluster"
```
framiere: When the rbac demo will get bigger, we could maybe leverage associative arrays like such:

```sh
resources=(
    'Group:connect-cluster::'
    'Topics:_secrets::(for secret Registry)'
    'Group:secret-registry::(for secret Registry)')

for resource in "${resources[@]}" ; do
    RESOURCE="${resource%%::*}"
    DESCRIPTION="${resource##*::}"

    echo -e "\n# Grant the principal User:$ADMIN_CONNECT to the ResourceOwner role for $RESOURCE $DESCRIPTION"
    echo "confluent iam rolebinding create --principal User:$ADMIN_CONNECT --role ResourceOwner --resource $RESOURCE --kafka-cluster-id $KAFKA_CLUSTER_ID"
    confluent iam rolebinding create --principal User:$ADMIN_CONNECT --role ResourceOwner --resource $RESOURCE --kafka-cluster-id $KAFKA_CLUSTER_ID
done
```

ybyzek: @framiere that's very cool bash mojo, but I intentionally didn't make those optimizations for the following reasons:

1. This demo is meant as a learning resource
2. For loops obscure all the required steps...all steps should be unfurled
3. Fancy bash makes it harder for bash newbies to parse
4. Optimizations should be made in a configuration tool

(In fact, I just unfurled my only for loop I had somewhere else)

ybyzek:
I believe I addressed all comments in this PR.  Please review/approve if it looks good

tobydrake7 (approved):
LGTM 👍

PR #197, opened 2019-07-10
Use explicit partition count for CUSTOMERS

rodesai:
In 5.3, KSQL changed the partition count for created sink topics to
be copied from the source topic rather than from the config file. This
patch updates the demo to use an explicit partition count so that ksql
can execute the join between customers and ratings.


ybyzek (approved):
@rodesai thanks!!

PR #203, opened 2019-07-16
DEVX-951: RBAC demo: move topic name to config file and use variable

ybyzek:


rspurgeon (approved):
LGTM

Very mintor observation: for files we source into the environment, we might want to follow (what I've seen as) more common nomenclature and give the file `.env` extension.   `local-demo.env` for example.  When I saw `.cfg` I went to go look and see if it was used by passing to a CP system for configuration.  I _think_ the former is more common and when I've seen `.env` in the past I know it's a system environment file.

ybyzek:
@rspurgeon -- thanks, as we discussed offline, I'll make that extension change in another PR.  Grazie!

PR #205, opened 2019-07-16
DEVX-956: RBAC: Auto-generate login.properties file from the configur…

ybyzek:
…ation file and rename the config file from .cfg to .env extension


security/rbac/scripts/enable-rbac-broker.sh
```diff
@@ -34,40 +34,40 @@ login_mds $MDS
...
+echo -e "\n# Grant principal User:$USER_ADMIN_SYSTEM the SystemAdmin role to the Kafka cluster"
+echo "confluent iam rolebinding create --principal User:$USER_ADMIN_SYSTEM --role SystemAdmin --kafka-cluster-id $KAFKA_CLUSTER_ID"
+confluent iam rolebinding create --principal User:$USER_ADMIN_SYSTEM --role SystemAdmin --kafka-cluster-id $KAFKA_CLUSTER_ID
 
-echo -e "\n# List the role bindings for User:$ADMIN_SYSTEM"
-echo "confluent iam rolebinding list --principal User:$ADMIN_SYSTEM --kafka-cluster-id $KAFKA_CLUSTER_ID"
-confluent iam rolebinding list --principal User:$ADMIN_SYSTEM --kafka-cluster-id $KAFKA_CLUSTER_ID
+echo -e "\n# List the role bindings for User:$USER_ADMIN_SYSTEM"
+echo "confluent iam rolebinding list --principal User:$USER_ADMIN_SYSTEM --kafka-cluster-id $KAFKA_CLUSTER_ID"
+confluent iam rolebinding list --principal User:$USER_ADMIN_SYSTEM --kafka-cluster-id $KAFKA_CLUSTER_ID
```
rspurgeon, line 47: One idea you might consider to reduce this duplication, here or in similar future projects, could be:
```
CMD="confluent iam rolebinding create --principal User:$USER_ADMIN_SYSTEM --role SystemAdmin --kafka-cluster-id $KAFKA_CLUSTER_ID"
echo $CMD
eval "$CMD"

CMD="confluent iam rolebinding list --principal User:$USER_ADMIN_SYSTEM --kafka-cluster-id $KAFKA_CLUSTER_ID"
echo $CMD
eval "$CMD"
```

security/rbac/scripts/create_login_properties.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python
```
rspurgeon, line 1: I'm not certain of the risks, but do you think it's worth being explict on the python version here?  Might prevent some oddities for users?
```suggestion
#!/usr/bin/env python3
```

rspurgeon (approved):
LGTM

I tested and works well, only a few minor comments.

security/rbac/scripts/enable-rbac-broker.sh
ybyzek: @rspurgeon thank you for this suggestion, 💯 agree.  I have filed DEVX-958 for tracking and will address throughout the entire script in a separate PR.

security/rbac/scripts/create_login_properties.py
ybyzek: @rspurgeon concerns about tying it to python3:

1. What if user has python2 but not python3?
2. The script doesn't do anything complicated and I tested both python2 and python3.

So leaving as is.

PR #206, opened 2019-07-17
DEVX-952: Remove deprecated (AK2.0) Connect internal converter configs

rspurgeon:
Prior to this change, the kafka connect `internal.value|key.converter` config values were required.  That has been deprecated by [KAFKA-5540](https://issues.apache.org/jira/browse/KAFKA-5540)

This change simply removes the config values from all files.  Will need to search upstream branches after merging for these in additional files


ybyzek (commented):
@rspurgeon LGTM!  Before approving, I just wanted to verify if you have validated at least one of these demos to ensure it still works?

PR #215, opened 2019-07-24
DEVX-952 Remove Connect Internal key/value converters

rspurgeon:
These config values have been deprecated: KAFKA-5540


ybyzek (approved):
LGTM!  Thanks for the PR and for validating the changes

PR #219, opened 2019-07-26
DEVX-988 - Add RST landing page for demos

joel-hamill:
https://confluentinc.atlassian.net/browse/DEVX-988
- related https://github.com/confluentinc/docs/pull/2411

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,10 @@
 */delta_configs/
 */target
 */*/target
+*/*/*/target
 img_data
 python_notify/
 .Python
 .vscode
 node_modules/
+.DS_Store
--- a/index.rst
+++ b/index.rst
@@ -1,0 +1,231 @@
+Demos
+=====
+
+This is a curated list of demos that showcase Apache Kafka® stream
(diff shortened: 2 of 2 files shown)


ybyzek:
@joel-hamill thanks for the PR.

A few clarification questions:

1. Is `index.rst` intended to be used only in docs.confluent.io?
2. Does this mean we have to make equivalent changes in two places now: index.rst and README.md, whenever something is updated in either?
3. Should the GH links mentioned index.rst point to `|release|` instead of `5.3.0-post`?

joel-hamill:
@ybyzek 
Is index.rst intended to be used only in docs.confluent.io?
> yes.  i'm open to other ideas?
Does this mean we have to make equivalent changes in two places now: index.rst and README.md, whenever something is updated in either?
> yes
Should the GH links mentioned index.rst point to |release| instead of 5.3.0-post?
> yeah that would make sense - except in that table (which i sloppily transferred as HTML, that won't work).

joel-hamill:
@ybyzek any more thoughts on this?

joel-hamill:
closing per discussion with @ybyzek

PR #221, opened 2019-07-29
DEVX-989 upgrade kafka-streams-example jar dependency

rspurgeon:
Prior to this change, 5.3.0-post branch would not operate (local or docker-compose)


gAmUssA (approved):
LGTM

PR #225, opened 2019-08-06
Fix docs build issues

joel-hamill:


ybyzek (approved):
LGTM, thanks @joel-hamill

PR #241, opened 2019-08-23
DEVX-921: Update cp-quickstart example to load new connectors with co…

ybyzek:
…nfluent CLI instead of curl


rspurgeon (approved):
LGTM 60+ lines gone... boom!

PR #243, opened 2019-08-26
add named C3 services + super user bindings

brianbushree:
### what
1. used C3 named services
2. bind super user to clusters 

### why
1. now we support multiple KSQLs/Connects in C3
2. be able to use the super-user for testing

### DIFF
--- a/security/rbac/rbac-docker/create-role-bindings.sh
+++ b/security/rbac/rbac-docker/create-role-bindings.sh
@@ -17,14 +17,14 @@ KSQL=ksql-cluster
 C3=c3-cluster
 
 SUPER_USER=professor
-SUPER_PASSWORD=professor
+SUPER_USER_PASSWORD=professor
 SUPER_USER_PRINCIPAL="User:$SUPER_USER"
 CONNECT_PRINCIPAL="User:fry"
 SR_PRINCIPAL="User:leela"
 KSQL_PRINCIPAL="User:zoidberg"
 C3_PRINCIPAL="User:hermes"
 
-XX_CONFLUENT_USERNAME=professor XX_CONFLUENT_PASSWORD=professor confluent login --url $MDS_URL
+XX_CONFLUENT_USERNAME=$SUPER_USER XX_CONFLUENT_PASSWORD=$SUPER_USER_PASSWORD confluent login --url $MDS_URL
 
 ################################### SETUP SUPERUSER ###################################
 echo "Creating Super User role bindings"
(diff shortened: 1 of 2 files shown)


ybyzek (approved):
@brianbushree -- LGTM!  I'd prefer to merge after @stan-confluent gives his 👍 as well.

security/rbac/rbac-docker/create-role-bindings.sh
```diff
@@ -120,7 +147,12 @@ echo "    connect cluster id: $CONNECT"
...
+echo "    super user account: $SUPER_USER_PRINCIPAL"
 echo "    connect service account: $CONNECT_PRINCIPAL"
 echo "    schema registry service account: $SR_PRINCIPAL"
 echo "    KSQL service account: $KSQL_PRINCIPAL"
-echo "    C3 service account: $C3_PRINCIPAL"
\ No newline at end of file
+echo "    C3 service account: $C3_PRINCIPAL"
+
+echo
+echo "To set service IDs as environment variables paste/run this in your shell:"
```
stan-is-hate, line 157: 👍

security/rbac/rbac-docker/docker-compose.yml
```diff
@@ -477,15 +477,21 @@ services:
...
       CONTROL_CENTER_MONITORING_INTERCEPTOR_TOPIC_PARTITIONS: 1
       CONFLUENT_METRICS_TOPIC_REPLICATION: 1
       PORT: 9021
+
+      # ========================= other services ==============================
+      # connect
+      CONTROL_CENTER_CONNECT_CONNECT1_CLUSTER: http://connect:8083
+      # ksql
+      CONTROL_CENTER_KSQL_KSQL1_URL: http://ksql-server:8088
+      CONTROL_CENTER_KSQL_KSQL1_ADVERTISED_URL: http://localhost:8088
```
stan-is-hate: I might be a little bit confused about how advertised urls work, will ksql advertise on localhost or ksql-server?

security/rbac/rbac-docker/create-role-bindings.sh
```diff
@@ -18,13 +18,40 @@ C3=c3-cluster
 
 SUPER_USER=professor
 SUPER_PASSWORD=professor
+SUPER_USER_PRINCIPAL="User:$SUPER_USER"
 CONNECT_PRINCIPAL="User:fry"
 SR_PRINCIPAL="User:leela"
 KSQL_PRINCIPAL="User:zoidberg"
 C3_PRINCIPAL="User:hermes"
 
 XX_CONFLUENT_USERNAME=professor XX_CONFLUENT_PASSWORD=professor confluent login --url $MDS_URL
```
stan-is-hate: Maybe use `SUPER_USER` in this line as well? (yeah, I should've done that originally lol)

stan-is-hate (approved):
Thanks Brian! Only minor comments from me, looks good overall!

security/rbac/rbac-docker/docker-compose.yml
brianbushree: the advertised URL is used by the frontend to talk directly to the service. so in this case, we want to use the port exposed on `localhost` since the browser runs outside the docker context

I'll add a comment about it in the setup

PR #244, opened 2019-08-27
error on missing dependencies & use docker ZK-shell

brianbushree:
### what
1. check dependencies before starting `openssl docker-compose docker confluent jq`
2. use docker zk-shell command

### why
1. fail fast 🙂 
2. don't require users to modify their path variable to include `zookeeper-shell`


ybyzek (approved):
LGTM, pending approval from @stan-confluent .  I haven't run the Docker/RBAC demo yet

stan-is-hate (approved):
Thanks Brian! Looks good!

PR #245, opened 2019-08-28
DEVX-1041: Fix ksql websocket initializor in rbac-docker

jonchiu:
rbac-docker example was using incorrect security handler. The fix was to set it to the same handler used by the rest servlet: InstallBearerOrBasicSecurityHandler


ybyzek (changes requested):
@jonchiu LGTM.  However, can we rebase/merge into 5.3.0-post instead of 5.3.x?

spena (approved):

jonchiu:
@ybyzek ping

PR #246, opened 2019-08-28
DEVX-963: RBAC: Add CSAS to KSQL example

spena:
Added a CSAS and a CTAS examples.


ybyzek (changes requested):
@spena thank you for the PR!  Can you please also update https://github.com/confluentinc/examples/blob/5.3.0-post/security/rbac/README.md#ksql with the new role bindings created?

PR #250, opened 2019-08-30
Added general rolebinding syntax

ElizabethSchafle:
Added some changes that would have been helpful to me when I started working with RBAC in the beginning, clarifying on the general syntax of defining and listing roles, as well as clarifying what roles and resources are available for ease of use.


ybyzek (changes requested):
Thanks for the PR @ElizabethSchafle .  Can you please place this at the end of the README, instead of in between KSQL & C3?

ElizabethSchafle:
> Thanks for the PR @ElizabethSchafle . Can you please place this at the end of the README, instead of in between KSQL & C3?

done

ybyzek (approved):

Issue #220, opened 2019-07-28
RBAC sample code error


manthak:
Downloaded the sample security code trying to run the sample code and having the following error

"kafka failed to log into your metadata server. please check all parameters and run again"

    only, NOT for production usage. https://docs.confluent.io/current/cli/index.html

Using CONFLUENT_CURRENT: /tmp/confluent.daFfsIy0
zookeeper is already running. Try restarting if needed
kafka is already running. Try restarting if needed
Sogin /apps/kafka/confluent-5.3.0/etc/kafka/server.properties /tmp/original_configs/server.properties ../delta_configs/server.properties.delta
Sleeping 5 seconds before login
Sleeping 5 seconds before login

# Logina http://localhost:8090   mds
./rbac_lib.sh: line 54: expect: command not found

Failed to log into your Metadata Server.  Please check all parameters and run again
[root@ftc-lbkfkdv204 scripts]#

ybyzek:
@manthak 

> ./rbac_lib.sh: line 54: expect: command not found

indicates that `expect` is required for this script to work.  Please install `expect` and try again.

Meanwhile, I'll file a jira to add a preflight check that `expect` is installed.

ybyzek:
PR submitted: https://github.com/confluentinc/examples/pull/253

PR #253, opened 2019-09-03
DEVX-994: Add a preflight check that expect is installed, for all rel…

ybyzek:
…evant demos where expect is used

Addresses https://github.com/confluentinc/examples/issues/220


rspurgeon (approved):
LGTM

PR #270, opened 2019-09-17
DEVX-1074: Update Docker-version of RBAC demo to use expect instead o…

ybyzek:
…f env variables


stan-is-hate (approved):
Thanks Yeva! Looks correct on surface, but I haven't used expect before, would be good to get a review from someone who had :)

ybyzek:
> Thanks Yeva! Looks correct on surface, but I haven't used expect before, would be good to get a review from someone who had :)

The code is already proven because the non-Docker version of the demo uses this exact code: https://github.com/confluentinc/examples/blob/5.3.1-post/security/rbac/scripts/rbac_lib.sh#L43-L61. And I validated it.

PR #273, opened 2019-09-23
DEVX-991 Update ACL Script documentation for CC Support

rspurgeon:


security/acls/acl.sh
```diff
@@ -168,8 +166,8 @@ ccloud service-account create $SERVICE_NAME --description $SERVICE_NAME || true
 SERVICE_ACCOUNT_ID=$(ccloud service-account list | grep $SERVICE_NAME | awk '{print $1;}')
 
 echo -e "\n# Create an API key and secret for the new service account"
-echo "ccloud api-key create --service-account-id $SERVICE_ACCOUNT_ID --cluster $CLUSTER"
-OUTPUT=$(ccloud api-key create --service-account-id $SERVICE_ACCOUNT_ID --cluster $CLUSTER)
+echo "ccloud api-key create --service-account-id $SERVICE_ACCOUNT_ID --resource $CLUSTER"
```
ybyzek: @rspurgeon for consideration, either:

1. Leave this to a different PR (DEVX-1084)
2. Update the script to check that the CLI version accepts `--resource` flag (requires enhancement to the `check_ccloud_v2` function)

.gitignore
```diff
@@ -1,12 +1,12 @@
-.idea/
-*/kafka-streams-examples
-*/delta_configs/
-*/target
-*/*/target
+**/.idea/
+**/kafka-streams-examples
```
ybyzek, line 2: What's the difference between single and double star?

security/acls/acl.sh
```diff
@@ -6,10 +6,7 @@
 ################################################################################
 #
 # Demo the new Confluent Cloud CLI and Access Control List (ACL) functionality
-# in your Confluent Cloud Enterprise cluster that has been enabled for ACLs
-#
-# ACLs are only available in Confluent Cloud Enterprise
-# ACLs are not available in Confluent Cloud Professional
+# in your Confluent Cloud cluster that has been enabled for ACLs
```
ybyzek:
```suggestion
# in your Confluent Cloud cluster
```

security/acls/acl.sh
```diff
@@ -123,11 +121,11 @@ if [[ $? != 0 ]]; then
   exit 1
 fi
 
-echo -e "\n# Verify cluster is a Confluent Cloud Enterprise cluster"
+echo -e "\n# Verify cluster supports ACLs"
```
ybyzek: Do we need this check any more?

ybyzek (changes requested):
Thanks for the PR @rspurgeon

security/acls/acl.sh
rspurgeon: preferred approach?

security/acls/acl.sh
rspurgeon: I'm not sure.  I don't know if it's possible for a cluster to exist that will fail this check across all environments.

security/acls/acl.sh
ybyzek: Actually, on thinking about it, my vote is BOTH (not either/or) for clarity and traceability.

So specifically, the suggestion would be:

1. In this PR, remove the change from `--cluster` to `--resource`.  Restore to `--cluster`
2. Submit a separate PR for the change from `--cluster` to `--resource`.
3. In that separate PR, augment `check_ccloud_v2` function to check the CLI version number...which will be useful down the road as well, to accommodate new CLI features and/or breaking changes

security/acls/acl.sh
ybyzek: Please double check with CCloud engineering team on this but IIUC, all ccloud clusters now support ACLs.

.gitignore
rspurgeon: maybe with `kafka-streams-examples` this isn't proper, not sure. With comman output directories like `target/` I wanted to ignore all `target/` folders regardless of location in the tree.

https://www.atlassian.com/git/tutorials/saving-changes/gitignore

ybyzek: Gotcha, thanks!

security/acls/acl.sh
ybyzek: https://docs.confluent.io/current/cloud/limits.html#cloud-limits

We should assume it's available on all clusters

rspurgeon: @ybyzek removed

ybyzek (approved):
Ship it @rspurgeon

PR #278, opened 2019-09-26
DEVX-557: Producer/consumer example to Confluent Cloud: Clojure

shayelkin:
Example of how to use Java interop to consume and produce from Confluent Cloud, based on the Scala examples.


shayelkin:
Unlike the Scala and Java examples though, I don't use Kafka's JsonSerializer and Deserializers, as the interop with the resulting `jackson` objects would make the code much more complex IMO, and contradicts the Clojure-y way to do that.

clients/cloud/clojure/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+```shell
+$ cat $HOME/.ccloud/config
+bootstrap.servers=<BROKER ENDPOINT>
+ssl.endpoint.identification.algorithm=https
+security.protocol=SASL_SSL
+sasl.mechanism=PLAIN
+sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required username\="<API KEY>" password\="<API SECRET>";
+
+# Quickstart
```
ybyzek: All the text in the README should be an exact replica of the other languages in this folder.  Can you please modify all the text to match?

clients/cloud/clojure/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+Once the above prerequisites are fulfilled, we can launch a Clojure REPL and start can start running the examples in this directory.
+
+   1. Run the consumer:
+   ```shell
+   $ cd examples/clients/clojd/clojure
+   $ lein consumer $HOME/.ccloud/config testtopic
```
ybyzek: For consistency this should also be `test1`

ybyzek (changes requested):
@shayelkin thanks for wrangling Clojure!

1. The text in the README should 💯 match the text in the other programming languages in this folder.  Can you please reformat?

2. In this PR, please also add a Clojure entry to the table here https://github.com/confluentinc/examples/tree/5.3.1-post/clients/cloud#confluent-cloud

clients/cloud/clojure/README.md
```diff
@@ -0,0 +1,72 @@
...
+   …
+   Polling
+   …
+   ```
+
+   2. In a seperate window, run the producer:
+   ```shell
+   $ cd examples/clients/clojd/clojure
```
ybyzek: Assume user is already in the proper directory (specifically `examples/clients/clojd`), no need to `cd`

clients/cloud/clojure/src/io/confluent/examples/clients/clj/producer.clj
```diff
@@ -0,0 +1,62 @@
...
+                        #(.send producer
+                                (ProducerRecord. topic k (json/write-str {:count %})))
+                        (range 6 11)))]
+          (.flush producer)
+          (while (not-every? future-done? futures)
+            (Thread/sleep 50))
+          (doseq [fut futures]
+            (try
+              (let [metadata (deref fut)]
+                (println (str "Produced record at " metadata)))
```
ybyzek: For consistency, this should read same as other languages, e.g. https://github.com/confluentinc/examples/blob/5.3.1-post/clients/cloud/nodejs/producer.js#L93

Same with consumer

mhowlett (approved):
lgtm after yeva's consistency issues fixed, but disclaimer, don't rely on my to know what's idiomatic and what's not, i've not written much clojure and not for a long time.

shayelkin:
Updated to match the Python example rather than the Scala one.

I'd argue that anything involving run from the command line and using `println` isn't too much Clojure-y: the usual method of development is tightly coupled to the REPL. 

That said, imo, the most modern Clojurey (as in post-1.7) thing to do would be to have transducers reading and writing to Kafka topics – but that, I think, is out of scope for a short example, but rather for a (small and useful) library.

clients/cloud/clojure/README.md
```diff
@@ -0,0 +1,79 @@
...
+Producing record: alice 	{"count":2}
+Producing record: alice 	{"count":3}
+Producing record: alice 	{"count":4}
+Produced record to topic test1 partiton [0] @ offest 0
+Produced record to topic test1 partiton [0] @ offest 1
+Produced record to topic test1 partiton [0] @ offest 2
+Produced record to topic test1 partiton [0] @ offest 3
+Produced record to topic test1 partiton [0] @ offest 4
+Producing record: alice 	{"count":5}
+Producing record: alice 	{"count":6}
```
shayelkin, line 49: (this differs from the Python example because there are two blocks of messages: one being sent with Callback, one by derefing the future, to demonstrate both methods)

ybyzek (approved):
@shayelkin I haven't run the example but approve based on the assumption that you have ;)  

Thank you again!

PR #279, opened 2019-09-26
DEVX-1118: Troubleshoot Leader showing "none"

ybyzek:


addisonhuddy (approved):
Works for me.  i think `docker container stop $(docker container ls -a -q) ; docker container rm $(docker container ls -a -q) ; docker volume prune -f` did the trick.

PR #285, opened 2019-09-30
DEVX-1131: Extend timeout in kafka-consumer-perf-test to prevent failures on run…

ybyzek:
…ning perf test

### DIFF
--- a/multiregion/README.md
+++ b/multiregion/README.md
@@ -260,7 +260,7 @@ Note: On failback from a failover to observers, any data that was not replicated
 
 ## Run end-to-end demo
 
-All the individual steps above can be run in on shot:
+All the individual steps above can be run with this automated script:
 
 ```
 ./scripts/start.sh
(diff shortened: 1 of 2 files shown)


multiregion/README.md
```diff
@@ -276,7 +276,7 @@ Stop the demo and all Docker containers.
 
 ## Troubleshooting
 
-1. If containers fail to ping each other (failures seen in running `./scripts/validate_connectivity.sh`), clean up the Docker environment before running the demo again.
+1. If containers fail to ping each other (e.g., failures seen in running `./scripts/validate_connectivity.sh`), clean up the Docker environment and then run the demo again.
```
rspurgeon: maybe suggest a command for "clean up the Docker environment"?

rspurgeon (approved):
Besides the one suggestion about providing docker help, LGTM

multiregion/README.md
ybyzek: @rspurgeon it's not rendering in the diff, but the command is shown right after this line

rspurgeon: doh! I see it now...   Now that I do, the commands do worry me a bit in the sense that they forecefully destroy docker resources across the board (including data volumes which might have value to the user).   Maybe we should be more specific about cleaning up only our resources by name or id to prevent accidental deletion of our end users resources?

ybyzek: @rspurgeon Agree!  I'll file a follow-on jira to filter on resources.

PR #287, opened 2019-10-01
DEVX-1135: Remove application.id: examples

ybyzek:


clickstream/docker-compose.yml
```diff
@@ -93,7 +93,6 @@ services:
       KSQL_BOOTSTRAP_SERVERS: kafka:29092
       KSQL_LISTENERS: http://0.0.0.0:8088
       KSQL_KSQL_SCHEMA_REGISTRY_URL: http://schema-registry:8081
-      KSQL_KSQL_SERVICE_ID: confluent_rmoff_01
```
rspurgeon, line 96: The referenced Jira discusses removal of application.id but i don't see anything about the removal of this value.  I don't have anything setup to test this so just want to verify you intend to remove _both_ the KSQL_APPLICATION_ID and KSQL_SERVICE_ID values?

ybyzek: Good observation, and yes, this is intentional.  In Slack we discussed that when those values are not overridden, they assume the default values.

> service.id has a default value (of `default_`) so for purposes of demos it’s fine to not set it (assuming only one KSQL cluster is being demoed).

rspurgeon (approved):
LGTM

clickstream/docker-compose.yml
vcrfxia: +1 to what @ybyzek said. It also doesn't hurt to set it, though, so perhaps we could set it to something like `clickstream_demo_` to call out that users will eventually want to set this value once they start using KSQL more seriously. (The service ID should end in an underscore for better readability of internal topic names.)

vcrfxia (approved):
LGTM. Thanks @ybyzek !

clickstream/docker-compose.yml
ybyzek: There's a whole bucket of defaults that these demos/examples don't override, that users may want to do in production, so it's not an anomoly.  We can evaluate this at a future time.

PR #288, opened 2019-10-02
DEVX-1137: Be more selective in which Docker resources are pruned

ybyzek:
Follow-up to comment in https://github.com/confluentinc/examples/pull/285

### DIFF
--- a/multiregion/README.md
+++ b/multiregion/README.md
@@ -278,7 +278,7 @@ Stop the demo and all Docker containers.
 
 ## Troubleshooting
 
-1. If containers fail to ping each other (e.g., failures seen in running `./scripts/validate_connectivity.sh`), then stop the demo, clean up the Docker environment, and restart the demo:
+1. If containers fail to ping each other (e.g., failures seen in running `./scripts/validate_connectivity.sh`), then stop the demo, clean up the Docker environment, and restart the demo.  If it still fails, restart Docker and run again.
 
 ```
 # Stop demo
--- a/multiregion/README.md
+++ b/multiregion/README.md
@@ -278,10 +278,18 @@ Stop the demo and all Docker containers.
 
 ## Troubleshooting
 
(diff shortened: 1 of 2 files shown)


multiregion/README.md
```diff
@@ -278,10 +278,18 @@ Stop the demo and all Docker containers.
...
 ```
+# Stop demo
 ./scripts/stop.sh
-docker container stop $(docker container ls -a -q) ; docker container rm $(docker container ls -a -q) ; docker volume prune -f
+
+# Clean up the Docker environment
+for c in $(docker container ls -q --filter "name=pumba"); do docker container stop "$c" && docker container rm "$c"; done
+docker-compose down -v --remove-orphans
+for v in $(docker volume ls -q --filter="dangling=true"); do docker volume rm "$v"; done
+docker volume prune -f
```
rspurgeon: If I understand correctly, line 290 will remove any volume that is dangling and line 291 will remove all unused volumes without any confirmation (forcefully).

Unless containers started outside of compose are creating volumes, I _think_ line 289 `docker-compose down -v` should be sufficient, and will only remove volumes related to this compose project instead of docker system wide.  If the `pumba` or other containers create volumes, then maybe we should name & remove them explicitly?

ybyzek:
@rspurgeon The reality is that we don't know why this sometimes fails, but these steps clear it up.

I removed the command that does a forced volume removal, and improved the stop script.

Leaving to the side the question of optimizing the commands, does this achieve the goal of not doing harm -- not removing anything that does not belong to the demo that shouldn't be removed?

rspurgeon:
> @rspurgeon The reality is that we don't know why this sometimes fails, but these steps clear it up.
> 
> I removed the command that does a forced volume removal, and improved the stop script.
> 
> Leaving to the side the question of optimizing the commands, does this achieve the goal of not doing harm -- not removing anything that does not belong to the demo that shouldn't be removed?

Yes, this looks sufficient to me.  IMO it's better to leave the docker system prune commands to manual operator invocation as they are so distructive.  LMK if you run into specific failures I could potentially help debug though.

rspurgeon (approved):
LGTM

PR #293, opened 2019-10-08
DEVX-1155: JMX CaughtUpReplicas

ybyzek:


jsancio (approved):
LGTM. Thanks!

multiregion/README.md
```diff
@@ -219,6 +219,23 @@ Observations:
 * In the first case, the consumer running in `east` reads from the leader in `west`, and so it is negatively impacted by the low bandwidth between `east` and `west`.  Its throughput is lower (e.g. `0.9025` MB.sec in the above example).
 * In the second case, the consumer running in `east` reads from the follower that is also in `east`. Its throughput is higher (e.g. `3.9356` MB.sec in the above example).
 
+## Monitoring Observers
```
jsancio, line 222: Note that we also have `IsNotCaughtUp` which returns `1` if all of the replicas (follower and observers) are not caught up to the leader. This is similar to under replicated. Users maybe interested on which partitions are not fully replicated (to all of the follower and observer) so that they can take some action.

cc @mumrah

ybyzek: @jsancio 

1. `IsNotCaughtUp` does not show up when running the following command:

```docker-compose exec broker-east-4 kafka-run-class kafka.tools.JmxTool --jmx-url service:jmx:rmi:///jndi/rmi://localhost:8094/jmxrmi```

Do we know why that is?

2. Proposal: let's leave that metric for the docs for now

jsancio: > Do we know why that is?
> 
>     1. Proposal: let's leave that metric for the docs for now

I need to investigate but I suspect that we don't have the latest version. `IsNotCaughtUp` was added after the preview was cut but it will be part of PA/GA.

https://github.com/confluentinc/ce-kafka/pull/711

ybyzek: 👍 that explains it.  I'll merge this PR as-is, and we'll revisit when a later image is available.

jsancio: Sounds good to me. Thanks for all of this great work!

PR #311, opened 2019-10-16
DEVX-1084 Update ccloud syntax cluster->resource, add version check

rspurgeon:


security/acls/acl.sh
```diff
@@ -48,6 +48,13 @@ check_timeout || exit 1
 check_mvn || exit 1
 check_expect || exit 1
 
+REQUIRED_CCLOUD_VER="0.185.0"
```
ybyzek: @rspurgeon excellent addition!  There are other demos that may need this so it may be more valuable to move this entire code segment directly INTO https://github.com/confluentinc/examples/blob/5.3.1-post/utils/helper.sh#L50-L60 .  WDYT?

ybyzek: See code function above on what it could look like: https://github.com/confluentinc/examples/blob/5.3.1-post/utils/helper.sh#L36-L48

rspurgeon: I added a fuller function but left a parameter w/ default in, in case a caller wants to check a against a different version

utils/helper.sh
```diff
@@ -59,6 +59,22 @@ function check_ccloud_v2() {
   return 0
 }
 
+function version_gt() { test "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1"; }
+
+function get_ccloud_version() {
+  ccloud version | grep "^Version:" | cut -d':' -f2 | cut -d'v' -f2
+}
+
+function check_ccloud_version() {
```
ybyzek, line 75: As a new function, it lost the check that `ccloud` exists in the first place.

If it doesn't exist, user gets error:


```
../../utils/helper.sh: line 65: ccloud: command not found

# Login
couldn't execute "ccloud": no such file or directory
    while executing
"spawn ccloud login --url https://stag.cpdev.cloud"
spawn ccloud login --url https://stag.cpdev.cloud
Failed to log into your cluster.  Please check all parameters and run again
```

utils/helper.sh
```diff
@@ -59,6 +59,22 @@ function check_ccloud_v2() {
...
+	REQUIRED_CCLOUD_VER=${1:-"0.185.0"}
+	CCLOUD_VER=$(get_ccloud_version)
+	if version_gt $REQUIRED_CCLOUD_VER $CCLOUD_VER; then
```
ybyzek, line 82: Nit: spacing is consistent with the rest of the file.

utils/helper.sh
rspurgeon: How do you want me to check?

Reusing this function:
https://github.com/confluentinc/examples/blob/5.3.1-post/utils/helper.sh#L28
Seems improper to me as that config file isn't required to operate ccloud anymore

ybyzek: Does this work? https://github.com/confluentinc/examples/blob/5.3.1-post/utils/helper.sh#L23-L26

rspurgeon: I added a function to check for the binary only and then referenced that from the existing check to not break existing function.  I then reference that in the version check.  Please see if that works.  I tested by removing the binary

ybyzek (approved):
LGTM, merge away!  Thanks @rspurgeon

PR #312, opened 2019-10-16
DOCS-1923 - Fix CCLOUD config syntax

joel-hamill:


ybyzek (changes requested):
Unfortunately the docs can't just change ``config`` for ``config.json`` in any of these commands.  To keep the commands as is, users have to manually create the ``config``file...or users have to use the `ccloud` CLI to do things like create topics.

PR #313, opened 2019-10-17
DEVX-1187: swap async and sync ordering in multiregion demo

ybyzek:


javabrett (approved):
Tested, LGTM, :shipit:

PR #314, opened 2019-10-17
DEVX-1186: Pumba network chaos containers: run longer, don't stop/exi…

javabrett:
…t after 30 minutes.

I don't know whether Pumba offers a "run indefinitely" mode - if it does, that would be better and clearer.  But 1000 days -> 24000h seems to be working.


ybyzek (approved):
@javabrett validated, LGTM.  I will merge now

PR #315, opened 2019-10-17
DEVX-1188: ccloud demo has old confluent syntax

ybyzek:


ccloud/docs/index.rst
```diff
@@ -428,13 +411,17 @@ Troubleshooting the demo
...
 
 
-2. To view log files, look in the current Confluent CLI temp director (requires demo to be actively running):
+2. To view log files, look in the current Confluent CLI temp directory (requires demo to be actively running):
 
    .. sourcecode:: bash
 
-        $ ls `confluent current | tail -1`
+        # View all files
+        $ ls `confluent local current | tail -1`
```
rspurgeon, line 419: Not related to this exact change, but for your future consideration... 

The confluent local warning is written to STDERR, so you can obfuscate the warning without having to use `tail` or other search function with:
`confluent local current 2> /dev/null` 

This _might_ be more reliable in various situations, not sure about here

rspurgeon (approved):
LGTM

ccloud/docs/index.rst
ybyzek: @rspurgeon thanks for the suggestion, I filed a separate jira to track that

PR #317, opened 2019-10-21
Remove extraneous port exposure

rmoff:
29092 is for internal (`broker`) access, and does not need to be exposed by Docker. Including it makes the Docker Compose less easy for people to understand.


ybyzek (approved):
@rmoff thanks for the PR, LGTM

PR #320, opened 2019-10-23
Multi-region replication demo: DEVX-1143: highlight delay in preferred leader election; DEVX-1221: Fix reassignment command that fails

ybyzek:


addisonhuddy (approved):
Works great.

PR #328, opened 2019-11-05
DEVX-1260 datagen connectors produce JSON values instead of AVRO

rspurgeon:


ybyzek (approved):
@rspurgeon  LGTM.  One thing to verify before merging though is {{value.converter.schemas.enable}} parameter setting...you probably did this already but it may help to test this change end-to-end (all the way through to CCloud).

rspurgeon:
> @rspurgeon LGTM. One thing to verify before merging though is {{value.converter.schemas.enable}} parameter setting...you probably did this already but it may help to test this change end-to-end (all the way through to CCloud).

Thanks @ybyzek , I did test the behavior all the way through, and I see 'schema JSON' all the way through.  I didn't check the default value for value.converter.schemas.enable

ybyzek:
Just to confirm, "all the way through" to KSQL queries?  Let's not merge until you validate KSQL queries work in Confluent Cloud.

rspurgeon:
> Just to confirm, "all the way through" to KSQL queries? Let's not merge until you validate KSQL queries work in Confluent Cloud.

Indeed there was an issue w/ processing these formats in KSQL.  Modified the PR so we generate raw JSON instead of the 'schemaed' version

PR #331, opened 2019-11-07
DEVX-1272: Reorder beginner ccloud demo so consumer reads from connect data

ybyzek:


README.md
```diff
@@ -9,6 +9,8 @@
 
 This is a curated list of demos that showcase Apache Kafka® event stream processing on the Confluent Platform, an event stream processing platform that enables you to process, organize, and manage massive amounts of streaming data across cloud, on-prem, and serverless deployments.
 
+<a href="http://www.youtube.com/watch?feature=player_embedded&v=muQBd6gry0U" target="_blank"><img src="http://img.youtube.com/vi/muQBd6gry0U/0.jpg" width="240" height="180" border="10" /></a>
```
rspurgeon: maybe look into making the demo preview larger and centered?  It feels a little slapped on in the current orientation IMO.

ccloud/README.md
```diff
@@ -20,15 +20,16 @@ This repo has a few resources to help you validate your solutions on Confluent C
 It steps through the following workflow.
 
 * Log in to Confluent Cloud
-* Create a demo environment and cluster, and specify them as the default
-* Create create a user key/secret
-* Create a Service Account and API key and secret
+* Create a new environment and specify it as the default
+* Create a new Kafka cluster and specify it as the default
+* Create a user key/secret pair
```
rspurgeon: Like the env, and cluster, do the creds get "selected" also?

For instance, in my log

```
specify active API key that was just created
ccloud api-key use ...
```

rspurgeon:
In my test, I didn't have docker running.  The demo goes really far along before it gets to local docker 'stuff'.   When it got there, it failed, but had already created lots of cloud stuff.  So I'm not sure the state of my environment now that the demo has failed.  Maybe we could avoid this some by checking for docker in the bootstrap steps of the demo instead of waiting until docker itself is used?

FYI: After this situation where docker wasn't running, I tried starting and rerunning the demo and got:

```
# Create and specify active environment
ccloud environment create demo-script-env
Error: 1 error occurred:
        * error creating account: Account name is already in use


Failed to create environment demo-script-env. Please troubleshoot and run again
```

README.md
ybyzek: * I made it a little bigger.
* I had tried for centering it but didn't figure it out.  Moving on ;)

ccloud/README.md
ybyzek: Yes!  Thanks

ybyzek:
@rspurgeon hopefully I addressed your feedback.  Please let me know what you think of the most recent copy

rspurgeon:
LGTM, nice work here

rspurgeon (approved):
LGTM

PR #333, opened 2019-11-07
DEVX-1273 update image provided by interactive team

rspurgeon:


ybyzek:
@rspurgeon not sure if it's too late to be asking this question, but is "consumers" the right word on the right?  For example, should it say "applications" instead?

rspurgeon:
> @rspurgeon not sure if it's too late to be asking this question, but is "consumers" the right word on the right? For example, should it say "applications" instead?

I agree the diagram isn't ideal, i'd take it back if I could.  I wrote it thinking about producers in one cluster with consumers in another thinking about active-passive hybrid in particular.  Fixing it will require rework from our internal team... LMK

ybyzek:
@rspurgeon  let's get this fixed before merging.  It's one word, hope is that it will be a relatively quick turnaround.

rspurgeon:
@ybyzek image has been updated and this PR rebased off 5.3.1-post

ybyzek (approved):
Awesome LGTM!

PR #336, opened 2019-11-08
DEVX-1276 disables loadBalancer by default in gke-base demo

rspurgeon:
This is required to run the demo w/ out setting the loadBalancer domain.

Thanks for catching @gemma-singleton


ybyzek (approved):
@rspurgeon I have not validated the change but am assuming you have, LGTM

gemma-singleton:
@ybyzek I just ran it with this change and it went up no problem

PR #341, opened 2019-11-12
DEVX-1274 Adds documentation for deploying JARs with Operator

rspurgeon:
Docs PR: #3179 contains the staged docs change


ybyzek (commented):
@rspurgeon should this explanation be added as well to `gke-base`?  Excellent explanations otherwise!

rspurgeon:
> @rspurgeon should this explanation be added as well to `gke-base`? Excellent explanations otherwise!

To just copy and paste it in there as well? Or a link instead, maybe from the 'Advanced' section?

ybyzek:
> To just copy and paste it in there as well? Or a link instead, maybe from the 'Advanced' section?

Since most of it is the same, can you turn it into an `include`?

rspurgeon:
> 
> Since most of it is the same, can you turn it into an `include`?

Done, staging docs updated

rspurgeon:
Do not merge, I have to remove the non-duplicate documentation first

rspurgeon:
@ybyzek I generalized the language and added an include, pulled it into both operator demos.  Please re-review, TY

kubernetes/docs/includes/deploy-jars-k8s.rst
```diff
@@ -0,0 +1,36 @@
...
+
+`Kafka Connect <https://docs.confluent.io/current/connect/index.html>`__ employs an extensible programming model to faciliate `Data Conversions <https://docs.confluent.io/current/connect/concepts.html#converters>`__, `Single Message Transformations (SMTs) <https://docs.confluent.io/current/connect/concepts.html#transforms>`__, or the development of custom connectors if a `supported connector <https://www.confluent.io/hub/>`__ is not available.  The Connect `packaging model <https://docs.confluent.io/current/connect/devguide.html#packaging>`__ will require that you deploy an archive containing your custom JAR to the `Kafka Connect's plugin path <https://docs.confluent.io/current/connect/userguide.html#connect-installing-plugins>`__.
+
+This demonstration utilizes a `custom Connector to generate mock events <https://github.com/confluentinc/kafka-connect-datagen>`__ which is **not** installed by default in the Confluent Kafka Connect Docker images.  In order to utilize this plugin, we build a custom Docker image which derives from the base Confluent Operator based Docker image and install our the custom connector using the ``confluent-hub`` client.
```
ybyzek:
```suggestion
This demonstration utilizes a `custom Connector to generate mock events <https://github.com/confluentinc/kafka-connect-datagen>`__ which is **not** installed by default in the Confluent Kafka Connect Docker images.  In order to utilize this plugin, we build a custom Docker image which is derived from the Confluent Operator based Docker image and install the custom connector using the ``confluent-hub`` client.
```

kubernetes/docs/includes/deploy-jars-k8s.rst
```diff
@@ -0,0 +1,36 @@
...
+
+Here is the ``Dockerfile`` snippet used to build these images (`source <https://github.com/confluentinc/kafka-connect-datagen/blob/0.1.x/Dockerfile-operator-local>`__):
```
ybyzek: To maintain a single source of truth, consider removing the snippet and just link to the Dockerfile itself.  That way future changes to the Dockerfile won't put this doc out of sync.

kubernetes/docs/includes/deploy-jars-k8s.rst
```diff
@@ -0,0 +1,36 @@
...
+
+Once the Docker image is built with your custom archive installed, Kubernetes will need to be able to pull this image from a Docker Registry to create the Pods.  The demonstration accomplishes this by publishing the image to `Docker Hub <https://hub.docker.com/r/cnfldemos/cp-server-connect-operator-with-datagen>`__.
+
+Finally your Operator Helm values will need to be updated to pull the custom images for your Pods.  This demo accomplishes this by overriding the ``connect`` image repository like so:
```
ybyzek:
```suggestion
Finally your Operator Helm values will need to be updated to pull the custom images for your Pods.  This demo accomplishes this by overriding the ``connect`` image to instead use the one published to Docker Hub as described in the previous paragraph:
```

kubernetes/docs/includes/deploy-jars-k8s.rst
```diff
@@ -0,0 +1,36 @@
...
+
+::
+
+  connect:
+  image:
+    repository: cnfldemos/cp-server-connect-operator-with-datagen 
+    tag: 0.1.6-5.3.1.0
+
+See the demo's `value.yaml <https://github.com/confluentinc/examples/blob/5.3.1-post/kubernetes/gke-base/cfg/values.yaml#L53>`__ file and the `kafka-connect-datagen <https://github.com/confluentinc/kafka-connect-datagen>`__ GitHub repository for the full example.
```
ybyzek: This will vary depending on the demo.  Consider either:

1. Adding a link to both
2. Removing the link from the include and add to each file below

ybyzek (commented):
@rspurgeon nice move to the include.  Some minor suggestions.

kubernetes/docs/includes/deploy-jars-k8s.rst
rspurgeon: It's the same across demos, demo 2 inherts demo 1.  This is how it was when the language was in demo 2 originally before including it across demo 1

kubernetes/docs/includes/deploy-jars-k8s.rst
rspurgeon: it would be great if i could link to a file in GitHub and have it rendered in RST, but I'm not aware of a solution.  I like having the Dockefile snippet here so a user can pull out the example right here and use in their own use.  IMO this is a low risk thing that drifts, and if it does it seems minimal risk vs the benefit to the reader to see a working example right away.

ybyzek: Sometimes that's true, and I'd say that's almost true in this case except this Dockerfile in particular is a bit overly customized -- no one is going to be able to copy/paste it as is because:

1. It uses env var like `${KAFKA_CONNECT_DATAGEN_VERSION}`
2. It assumes locally compiled datagen connector with the `COPY` command
3. As a result of #2, the confluent-hub command is from local instead of Confluent Hub

We actually don't want people to use that model, it's a more advanced workflow.

Ideally we would want to steer people to the easier workflow where it's just:

```
RUN confluent-hub install --no-prompt kafka-connect-datagen:0.1.6
```

Therefore, it may in fact do more harm than good to include the whole Dockerfile.

kubernetes/docs/includes/deploy-jars-k8s.rst
ybyzek: My bad.

kubernetes/docs/includes/deploy-jars-k8s.rst
rspurgeon: IMO The Dockerfile should be parsable by a novice Dockerfile builder.  Variables are common in Dockefiles.  This whole section is about _customzied_ jars/connectors for SMT, transformations, etc... in those cases confluent hub will not help them.  Linking them to the same solution doesn't help them if they can't read it there either.

I suggest we either remove the example and don't link them or spend the time making a generic one, which may exist in the confluent docs somewhere already.

rspurgeon: We could just link out to the confluent-hub reference document, and remove the example entirely but link to the kafka-conenct-datagen repo generally instead of the specific Dockerfile if we don't think it would be helpful (I think the snippet is helpful).

https://docs.confluent.io/current/connect/managing/confluent-hub/command-reference/index.html

ybyzek (approved):
LGTM

PR #342, opened 2019-11-13
DEVX-1284: For pre-5.4 GA demos, pull in specific version of kafka-co…

ybyzek:
…nnect-datagen

I have validated one of the demos with this PR, specifically `cp-quickstart/start.sh`.

Related to future change: https://github.com/confluentinc/kafka-connect-datagen/pull/44

### DIFF
--- a/ccloud/beginner-cloud/Dockerfile
+++ b/ccloud/beginner-cloud/Dockerfile
@@ -17,4 +17,4 @@ FROM confluentinc/cp-kafka-connect:5.3.1
 
 ENV CONNECT_PLUGIN_PATH="/usr/share/java,/usr/share/confluent-hub-components"
 
-RUN confluent-hub install --no-prompt confluentinc/kafka-connect-datagen:latest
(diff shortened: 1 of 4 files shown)


C0urante:
@ybyzek the changes look acceptable, but have we considered using a substitution so that if the version needs to be adjusted in the future we'll only have to tweak one line and the chances of accidentally missing an occurrence of the version will be eliminated?

ybyzek:
@C0urante 💯 agree with the suggestion.  But it'll take careful thought about how to implement this across all demos in the repo.  This suggestion will not be implemented in this PR but I have filed DEVX-1290 to track.  Thank you!

C0urante (approved):
LGTM

I'm a little embarrassed to admit it... but I thought this PR was for the docs repo, which already has its own RST substitution logic. Yeah, let's definitely save adding substitution logic to this repo for a separate issue/PR 😅

rspurgeon (approved):
LGTM, tested locally

PR #343, opened 2019-11-15
DEVX-1295: Detection CP vs CCS is broken

ybyzek:
- Fixed case
- Replaced `is_ce` with `is_cp`
- Added missing double dash

### DIFF
--- a/utils/helper.sh
+++ b/utils/helper.sh
@@ -170,16 +170,19 @@ function check_running_cp() {
   return 0
 }
 
-function is_cp() {
-  type=$( confluent local version 2>/dev/null | tail -1 | awk -F: '{print $1;}')
-  if [[ "$type" == "Confluent Platform" ]]; then
-    return 0
-  elif [[ "$type" == "Confluent Community Software" ]]; then
-    return 1
-  else
-    echo -e "\nCannot determine if Confluent Platform or Confluent Community Software from `confluent local version`. Assuming Confluent Community Software\n"
-    return 1
-  fi
+function check_cp() {
+  type=$( confluent local version 2>/dev/null | tail -1 | awk -F: '{print $1;}' | tr '[:lower:]' '[:upper:]')
+  case $type in
+    *PLATFORM*)
+      return 0 ;; 
+    *COMMUNITY*)
+      return 1 ;;
+    *)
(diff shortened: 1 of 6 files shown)


utils/helper.sh
```diff
@@ -170,14 +170,14 @@ function check_running_cp() {
...
     return 0
-  elif [[ "$type" == "Confluent Community software" ]]; then
+  elif [[ "$type" == "Confluent Community Software" ]]; then
     return 1
   else
-    echo -e "\nCannot determine if Confluent Platform or Confluent Community software from `confluent local version`. Assuming Confluent Community\n"
+    echo -e "\nCannot determine if Confluent Platform or Confluent Community Software from `confluent local version`. Assuming Confluent Community Software\n"
     return 1
   fi
 }
```
rspurgeon, line 186: @ybyzek 100% for your consideration but thought I would suggest another version of `is_cp` that _may_ be more tollerant to changes in the textual outputs from cnfl engeineering

```
function is_cp() {
  type=$( confluent local version 2>/dev/null | tail -1 | awk -F: '{print $1;}' | tr '[:lower:]' '[:upper:]' )
  rv=1
  case $type in
    *PLATFORM*)
      rv=0
      ;;
    *)
      echo -e "\nCannot determine if Confluent Platform or Confluent Community Software from `confluent local version 2>/dev/null`. Assuming Confluent Community Software\n"
      ;;
  esac
  return $rv
}
```

Alternatively/additionally you could search for COMMUNITY and print a specific error, but the default will be err code 1 if it can't determine the type of installation and print out details.   I believe the use of `tr` instead of shell options to disable character casing is better as it's not standard across shells AFAIK.

rspurgeon (approved):
LGTM Tested locally, and added a suggestion for your consideration but not a blocker

utils/helper.sh
ybyzek: @rspurgeon good suggestion!  I've modified the code in the latest commit.  I'd appreciate your review of the changes.

Also, I changed the name from `is_` to `check_` because `is_` implies boolean (which it's not), the logic is actually evaluating exit codes.

rspurgeon:
Still LGTM @ybyzek

PR #346, opened 2019-11-19
DEVX-1227: update links with UTM codes

ybyzek:


rspurgeon (approved):
LGTM, I didn't test each URL, but I scanned the individual files for consistency in the utm values

eunjinoh (approved):
Since I am not too familiar with utm parameter implementation, i could only review the values. The parameter values look good to me.

PR #354, opened 2019-11-22
DEVX-1311: Rename replicator-security yml files

ybyzek:


rspurgeon (approved):
LGTM, this is a nice change and more idiomatic compose usage.  I tested all 7 configurations with no issue.

ybyzek:
@thomaskwscott please let me know if there are any concerns with this PR, thanks in advance.

PR #355, opened 2019-11-22
REP-499 added replicator rest extension integration to the multi-dc demo

thomaskwscott:
This PR adds Replicator monitoring to the multi-dc demo. 

Right now the replicator docker images hardcode the classpath to /etc/kafka-connect/jars/ so some extra containers are necessary. This will be addressed in the image post 5.4.x


ybyzek (commented):
@thomaskwscott thank you for the PR!

As we discussed, I will review this PR in tandem with the docs PR that enhances https://docs.confluent.io/current/installation/docker/installation/replicator.html , whenever that second PR is available

multi-datacenter/docker-compose.yml
```diff
@@ -115,6 +115,8 @@ services:
     ports:
       - "8382:8382"
       - "9892:9892"
+    volumes:
+      - mi3:/etc/kafka-connect/jars/
```
ybyzek: @thomaskwscott this container's Docker image is already using `confluentinc/cp-enterprise-replicator`, so it's not clear why the PR uses `replicator-for-jar-transfer`?  (same with the connect container below)

As a result, this is in the logs:

```
multi-datacenter(thomaskwscott-5.4.x) ✗: docker-compose logs replicator-for-jar-transfer 
Attaching to replicator-for-jar-transfer
replicator-for-jar-transfer    | mkdir: cannot create directory ‘/opt/kafka-connect-replicator’: File exists
```

ybyzek (changes requested):
@thomaskwscott this PR doesn't appear to work as-is in my environment, using the latest 5.4.x images that I pulled this morning.  Did this work in your environment, and if yes, using which images?

1. I checked out the PR and the run fails:

```
connect-dc2                    | org.apache.kafka.connect.errors.ConnectException: Failed to find any class that implements interface org.apache.kafka.connect.rest.ConnectRestExtension and which name matches io.confluent.connect.replicator.monitoring.ReplicatorMonitoringExtension
```

2. See comment about transferring jars.

multi-datacenter/docker-compose.yml
thomaskwscott: We need to put the rest extension jar on the Connect worker classpath. All the images hardcode the classpath here: https://github.com/confluentinc/cp-docker-images/blob/5.3.1-post/debian/enterprise-replicator-executable/include/etc/confluent/docker/launch#L41 
There's no easy workaround to move the Replicator jar into the classpath directory when the container starts up as it has to be done before the connect worker is launched. 

The mount solution remounts, just the rest extension jar in the classpath dir (we can't have all of Replicator in there)

thomaskwscott:
It works in my environment with images from ECR last week, will check with latest

multi-datacenter/docker-compose.yml
ybyzek: @thomaskwscott we should avoid this extra Docker container and instead implement a cleaner solution, if possible.  Here are two ideas:

1. Why wouldn't this work in the same manner as https://github.com/confluentinc/cp-demo/blob/5.3.1-post/docker-compose.yml#L169

2. If it doesn't work, a workaround: copy the jar to the file and then let the Docker container continue (e.g. example https://github.com/confluentinc/cp-demo/blob/5.3.1-post/docker-compose.yml#L32)

thomaskwscott:
@ybyzek I removed the unnecessary mkdir and confirmed it's working on the latest images. Can you retest?

multi-datacenter/docker-compose.yml
thomaskwscott: I trued with the approach from 1 initially, that's how I found the hard coded value. I suspect this is not doing anything in cp-demo and won't work here. 
The 2 approach is nice, will try it today.

ybyzek: @thomaskwscott IIUC, you are saying that this line is a noop: https://github.com/confluentinc/cp-demo/blob/5.3.1-post/docker-compose.yml#L169 ?

That doesn't appear to be the case.  Here's a test, I changed CLASSPATH to `foobar`, and the result is it gets set in the Docker container:

```
$ docker-compose exec connect env | grep CLASSPATH
CLASSPATH=foobar
CUB_CLASSPATH=/etc/confluent/docker/docker-utils.jar
```

So can you try by setting the `CLASSPATH` to include the required jar?  That would be a simpler method than overriding the `command` statement.

ybyzek: As we discussed in Slack:

1. cp-demo uses that ENV CLASSPATH setting for running kafka-console-producer/consumer commands from CLI
2. @thomaskwscott will raise a jira to augment the Docker image to allow CLASSPATH to be overridden so we don't have to do the command hack (please post jira here)

multi-datacenter/docker-compose.yml
```diff
@@ -115,6 +115,7 @@ services:
     ports:
       - "8382:8382"
       - "9892:9892"
+    command: "bash -c 'cp /usr/share/java/kafka-connect-replicator/replicator-rest-extension-5.4.0-SNAPSHOT.jar /etc/kafka-connect/jars/; /etc/confluent/docker/run'"
```
ybyzek: @thomaskwscott Hard-coding this to `5.4.0-SNAPSHOT` will require manual edit for each release bump.  Instead, can you please modify this to catch any release?  something along the lines of `...replicator-rest-extension-*.jar...`

cc: @rspurgeon

ybyzek:
@thomaskwscott my last feedback is https://github.com/confluentinc/examples/pull/355#discussion_r353328629 and then I think the PR will be good to go (once I validate one last time).

multi-datacenter/docker-compose.yml
thomaskwscott: Done

ybyzek (approved):
Validated and LGTM, thank you @thomaskwscott !

PR #362, opened 2019-11-28
Minor fixes to elasticsearch index configs and Kibana dashboard 

sarwarbhuiyan:
1. Added control center
2. Fixed Elasticsearch types to be "_doc" rather than the expression "type.name=kafkaconnect"
3. Fixed Kibana visualizations which were out of alignment with data (FULL_NAME vs FIRST_NAME and LAST_NAME)


ybyzek:
@sarwarbhuiyan thanks for the PR!  Can you add a note into `postgres-debezium-ksql-elasticsearch-docker-short.adoc` that highlights for users that they can use C3 UI for management and monitoring? (also consider adding at least one screenshot).  Otherwise, they may not realize that it's part of the Docker environment.

@rspurgeon since you are validating this same demo for 5.4.0, would you also be able to validate the changes in this PR?

rspurgeon (approved):
LGTM, @sarwarbhuiyan I will merge this down and forward

PR #369, opened 2019-12-04
MINOR: Update rbac-demo configs as per changes in 5.4.x

omkreddy:


security/rbac/delta_configs/kafka-rest.properties.delta
```diff
@@ -1,3 +1,4 @@
+zookeeper.connect=localhost:2181
```
omkreddy, line 2: looks like this was missing in 5.3

omkreddy:
@ybyzek  @milosimpson @rajinisivaram
I ran the demo on local install of Confluent Platform. I have not tried with docker install.

rajinisivaram (commented):
@omkreddy Thanks for the PR, LGTM. Since you are here, could you also change the listener name in `server.properties.delta` from RBAC to TOKEN please? We can do it in a separate PR if that is easier.

security/rbac/delta_configs/kafka-rest.properties.delta
ybyzek: @omkreddy IIUC, `zookeeper.connect` is not required because the properties file already has `bootstrap.servers=localhost:9092` (see below).  Plus the demo works without this line (at least it did in `5.3.1-post`).  Can you please clarify/confirm, and if so, remove this line?

ybyzek (commented):
@omkreddy thank you for the PR

security/rbac/delta_configs/kafka-rest.properties.delta
omkreddy: @ybyzek kafka-rest still depends on ZK for licence validation. So we require `zookeeper.connect` config when used with confluent security extensions. kafka-rest server startup will fail without this config. 

> @omkreddy Thanks for the PR, LGTM. Since you are here, could you also change the listener name in `server.properties.delta` from RBAC to TOKEN please? We can do it in a separate PR if that is easier.

@rajinisivaram Thanks for the review. updated the PR.

rajinisivaram (approved):
@omkreddy Thanks for the updates, LGTM

security/rbac/delta_configs/kafka-rest.properties.delta
ybyzek: @omkreddy thanks for the explanation.  The default must already be `localhost:2181` because the demo works without setting that value.

Given that it's required but the default works, can you please add a comment above the line, something like: `# Configure connection to ZooKeeper, required for license validation of the Confluent security extensions`?

ybyzek (approved):
LGTM, thank you!

omkreddy:
@ybyzek  Thanks for the review.  I don't have merge permission. Please merge the PR when you get a chance.

PR #370, opened 2019-12-04
Remove duplicated KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR

antonmry:
It works but some IDEs report it as an error


ybyzek (approved):
@antonmry thanks for the PR, approved.

PR #376, opened 2019-12-06
DEVX-1318: Multi Region Replication demo -- spruce up images

ybyzek:


rspurgeon (approved):
LGTM, links render properly... nice

PR #379, opened 2019-12-06
Update control-center-dev.properties.delta to match the latest names

srpanwar-confluent:


srpanwar-confluent:
Closing this as there is already another PR with has this fix included

PR #384, opened 2019-12-13
DEVX-1331 Validates microservices-orders to 5.4.x

rspurgeon:


microservices-orders/get-kafka-streams-examples.sh
```diff
@@ -2,7 +2,7 @@
 
 # Compile java client code
 [[ -d "kafka-streams-examples" ]] || git clone https://github.com/confluentinc/kafka-streams-examples.git
-(cd kafka-streams-examples && git fetch && git checkout 5.4.0-post && git pull && mvn clean compile -DskipTests package)
+(cd kafka-streams-examples && git fetch && git checkout 5.4.x && git pull && mvn clean compile -DskipTests package)
```
rspurgeon, line 5: @ybyzek not complete sure about this... there is no 5.4.0-post as of yet, so we'll have to coordinate a change when release cut occurs, otherwise this can't be ran.  Or, is it valid to leave it at 5.4.x going forward?

ybyzek: It depends.  

@mohnishbasha -- will this line be automatically changed to `5.4.0-post` when the RC is cut?  Or should we manually change it to `5.4.0-post` now?

ybyzek (changes requested):
@rspurgeon in my environment:

pass: running `./start.sh`
fail: running `docker-compose up -d --build`

Essentially the Docker version of the demo fails in my environment, potentially due to subtle timing changes in the new images.  Does the Docker version of the demo work in your environment?

rspurgeon:
@ybyzek 
> Does the Docker version of the demo work in your environment?

Yes, it appears to work.  I can verify w/ the `read-topics-docker.sh` script.  If memory serves I verified w/ the docker solution more so than the local installation.  I'll run through both again to re-verify

rspurgeon:
> pass: running `./start.sh`
> fail: running `docker-compose up -d --build`

@ybyzek i've verified this twice, with both, docker and local using the most recent builds.  Can you send over an error from the failure you're seeing?

rspurgeon:
@ybyzek Updated w/ a new cub parameter for timing, and force pushed as I've rebased off 5.4.x

ybyzek (approved):
LGTM, ship it!

PR #387, opened 2019-12-16
DEVX-1384: Uses "latest" version of GKE during cluster creation

rspurgeon:
This function was unknown to me prior.  This enables the usage of latest for anyone running the demo create cluster function.  Prior to this, the versions would become outdated very quickly and users would see errors


ybyzek (approved):
LGTM

PR #388, opened 2019-12-16
DEVX-1385 corrects version for jackson xml core lib

rspurgeon:
This makes the jackson version inherited from the parent pom (modeled after kafka-streams-examples) and results in the proper downloading of the dependency and resolves the ClassNotFound exception that was at the core of this issue.

Tested locally against stag Confluent Cloud

### DIFF
--- a/clients/cloud/java/pom.xml
+++ b/clients/cloud/java/pom.xml
@@ -102,7 +102,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>[2.8.11.1,)</version>
+            <version>${jackson.version}</version>
         </dependency>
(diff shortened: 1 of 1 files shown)


ybyzek (approved):
@rspurgeon thanks for the quick fix!

1. I validated that this fix works and the CCloud/CLI example succeeds.  LGTM.

2. There are references to the Jackson version in the Kotlin and Scala code examples as well.  Do those need to be updated? (If so, a separate jira and PR is ok)

```
clients/cloud/kotlin/build.gradle.kts:  compile("com.fasterxml.jackson.core:jackson-databind:[2.8.11.1,)")
clients/cloud/scala/build.sbt:      "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.5",
```

PR #393, opened 2019-12-19
DEVX-1399 makes ES and Kibana check functions verify by port only

rspurgeon:
Also updated the es script to only invoke curl once, check for curls existence
and cleaned up a little of the conditional logic

### DIFF
--- a/utils/helper.sh
+++ b/utils/helper.sh
@@ -186,25 +186,21 @@ function check_cp() {
 }
 
 function check_running_elasticsearch() {
-  expected_version=$1
+  check_curl || exit 1
+  check_jq   || exit 1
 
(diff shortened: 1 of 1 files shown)


ybyzek (approved):
LGTM, assuming you have validated for both tarball and Docker?

rspurgeon:
> LGTM, assuming you have validated for both tarball and Docker?

I tested w/ bash only.   The scripts don't run anywhere but locally in a bash shell, and also just check network ports, so there _shouldn't_ be any difference.  But I can hold off on merging and verify with a few of the demos.

ybyzek:
> The scripts don't run anywhere but locally in a bash shell

Good catch, it's for tarball only.  Ship it!

PR #396, opened 2019-12-20
DEVX-1412 makes grafana check network only

rspurgeon:
This is part of a larger effort to improve the services checkes and does
resolve the jira entirely.

This changes to use the grafana api route which includes json and the version


ybyzek (approved):
LGTM

PR #397, opened 2019-12-20
quiets check_running_cp a bit to reduce demo noise

rspurgeon:
the cp local warning gets very redundant in demo scripts.  This function is just meant to check the version of the installed cp, so don't we need to show the warning here

### DIFF
--- a/utils/helper.sh
+++ b/utils/helper.sh
@@ -161,9 +161,9 @@ function check_running_cp() {
   check_curl
   expected_version=$1
 
-  actual_version=$( confluent local version | tail -1 | awk -F':' '{print $2;}' | awk '$1 > 0 { print substr($1,1,3)}' )
+  actual_version=$( confluent local version 2>/dev/null | awk -F':' '{print $2;}' | awk '$1 > 0 { print substr($1,1,3)}' )
   if [[ $expected_version != $actual_version ]]; then
-    echo -e "\nThis script expects Confluent Platform version $expected_version but the running version is $actual_version.\nTo proceed please either: change the examples repo branch to $actual_version or update the running Confluent Platform to version $expected_version.\n"
+    printf "\nThis script expects Confluent Platform version $expected_version but the running version is $actual_version.\nTo proceed please either: change the examples repo branch to $actual_version or update the running Confluent Platform to version $expected_version.\n"
     exit 1
   fi
 
--- a/utils/helper.sh
+++ b/utils/helper.sh
@@ -157,8 +157,17 @@ function check_gcp_creds() {
   fi
 }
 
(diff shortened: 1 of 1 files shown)


utils/helper.sh
```diff
@@ -161,9 +161,9 @@ function check_running_cp() {
   check_curl
   expected_version=$1
 
-  actual_version=$( confluent local version | tail -1 | awk -F':' '{print $2;}' | awk '$1 > 0 { print substr($1,1,3)}' )
+  actual_version=$( confluent local version 2>/dev/null | awk -F':' '{print $2;}' | awk '$1 > 0 { print substr($1,1,3)}' )
```
ybyzek, line 173: If `confluent local version` fails, this would cause it fail silently.  `confluent` was pulled out of CP 5.3, it's coming back in 5.4, but it's something to take into consideration.

utils/helper.sh
```diff
@@ -161,9 +161,9 @@ function check_running_cp() {
...
   if [[ $expected_version != $actual_version ]]; then
-    echo -e "\nThis script expects Confluent Platform version $expected_version but the running version is $actual_version.\nTo proceed please either: change the examples repo branch to $actual_version or update the running Confluent Platform to version $expected_version.\n"
+    printf "\nThis script expects Confluent Platform version $expected_version but the running version is $actual_version.\nTo proceed please either: change the examples repo branch to $actual_version or update the running Confluent Platform to version $expected_version.\n"
```
ybyzek, line 175: For my own understanding, why switch from `echo -e` to `printf`?

ybyzek (commented):
Minor questions

utils/helper.sh
rspurgeon: `echo -e` was rendering awkwardly in my shell (zsh) and it seems printf is more portable and recommended and preferred with the escape sequences.

Good explanation here: https://askubuntu.com/a/537987

utils/helper.sh
rspurgeon: when you say `confluent local version` fails, what scenarios are you contemplating specifically (so I can test this change)?

ybyzek: `confluent` bin isn't found?

rspurgeon: I added a new function to make this scenario very clear, see latest please.

I tested thus:

```
sh -c "source utils/helper.sh;PATH=/usr/bin;check_running_cp 5.4"

confluent command not found.  Please check your Confluent Platform installation
```

```
sh -c "source utils/helper.sh;check_running_cp 5.4;echo $?"
0
```

ybyzek (approved):
LGTM

PR #399, opened 2019-12-30
Update to 5000 records

AdheipSingh:
Num records to 5000


ybyzek:
Intentionally set to 200 because of latency.

PR #400, opened 2019-12-30
Update consumer

AdheipSingh:
Consumer config missing, added it for west region.


davetroiano:
The proposed change isn't desired. The first consumer perf test shows non-MRC throughput ("reading from Leader in west") for comparison to what you see in the second test when fetching from observers, so we don't actually want to set `client.rack`

PR #408, opened 2020-01-16
Fix 5.4.0 post-release docker image tags

andrewegel:
As per https://confluent.slack.com/archives/C09EP1SS3/p1579053207364100

### DIFF
--- a/ccloud/beginner-cloud/docker-compose.yml
+++ b/ccloud/beginner-cloud/docker-compose.yml
@@ -2,7 +2,7 @@
 version: '2'
 services:
   connect-cloud:
-    image: cnfldemos/cp-server-connect-datagen:0.1.7-5.4.x-latest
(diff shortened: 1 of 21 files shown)


ccloud/beginner-cloud/docker-compose.yml
```diff
@@ -2,7 +2,7 @@
 version: '2'
 services:
   connect-cloud:
-    image: cnfldemos/cp-server-connect-datagen:0.1.7-5.4.x-latest
+    image: cnfldemos/cp-server-connect-datagen:0.1.7-5.4.0
```
andrewegel, line 5: As per slack convo:

```
we ( I ), need to manually build these
```

rspurgeon: we don't need to hold up the PR for this, will take me a bit to build and publish them

PR #410, opened 2020-01-16
DEVX-1379: Use 5.4.0-standalone built jar for kafka-streams

rspurgeon:
kafka-streams-examples 5.4.0-post is available, and it doesn't build w/ the SNAPSHOT on that branch, so this fixes that routine up in microservices and music


ybyzek (approved):
LGTM

PR #415, opened 2020-01-21
DEVX-1290: Create fixed variables tied to CP release for consistency across demos

ybyzek:
@rspurgeon it's still undecided if we want to implement this for all demos, but for now we can clean up the implementation in the RBAC Docker demo so we have a good model.  So this PR is just for RBAC Docker demo for now.


rspurgeon (approved):
LGTM, tested locally

PR #417, opened 2020-01-22
DEVX-1448: Add config parameter for release as used in start.sh

ybyzek:


utils/config.sh
```diff
@@ -0,0 +1,5 @@
+# File with demo configuration parameters
+
+export CP_VERSION_MAJOR=5.4
+export CP_VERSION_FULL=5.4.0
```
rspurgeon: Would it make sense to derive the major version from the full version to reduce the various required to be managed?

Something like:
`CP_VERSION_MAJOR=${CP_VERSION_FULL%.*}`

A little harder w/ the branch but above maybe saves a substitution

kinesis-cloud/start.sh
```diff
@@ -3,10 +3,14 @@
 # Source library
 . ../utils/helper.sh
 
+# Source Confluent Platform versions
+. ../utils/config.sh
```
rspurgeon: what do you thnk about sourcing `config.sh` inside `helper.sh` that way helper.sh gets this by default, then where we don't already source in helper we just source that in which will be standard across most of the scripts, instead of having to source in multiple scripts

utils/config.sh
```diff
@@ -0,0 +1,5 @@
...
+export GH_BRANCH=5.4.0-post
```
rspurgeon: we might want to also consider adding something like `CP_JAR_VERSION=5.4.0`, which on `-x` branches could be similar to `5.4.0-SNAPSHOT`, or similar idea to this.  Possibly, in this change or later one, derive CP_JAR_VERSION's value entirely from the other variables.  We need a value like this in `microservices-orders/services-start.sh` and maybe elsewhere

utils/config.sh
ybyzek: @rspurgeon A concern with doing this is I don't remember how previews are managed.  For example, the 5.4 preview (pre-GA).  The MAJOR would be 5.4 (and that's what Confluent CLI would show) but I don't remember what the full would show (as used in other commands).  So while I acknowledge that this optimization reduces the number of lines to manage, it opens potential problems later.  tl;dr the value in keeping them separate is future-proofing against mis-matched naming conventions that we don't know.

kinesis-cloud/start.sh
ybyzek: @rspurgeon while it requires an extra source, they are conceptually different things.  One is a library of functions, one is a configuration file.  I don't have a strong opinion to keep them as separate, but just wanted to highlight the reasoning for doing so.  WDYT?

utils/config.sh
```diff
@@ -0,0 +1,5 @@
+# File with demo configuration parameters
+
+export CP_VERSION_MAJOR=5.4
```
rspurgeon: Is `export` necessary here?

kinesis-cloud/start.sh
rspurgeon: IMO the value of keeping them in a separate file is just so that tooling automation can update the file easily and prevent source control conflict issues.   I think the values in config.sh should be global to all our demos, so I don't see the need to require us to source in multiple script files when helper can just pull it in itself.  Or do it vice versa if it's the naming that's not right

utils/config.sh
ybyzek: Good idea.  In latest commit I added JAR_VERSION.

kinesis-cloud/start.sh
ybyzek: Addressed as you suggested.

utils/config.sh
ybyzek: It appears not.  Now removed.

microservices-orders/services-start.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
 
+# Source Confluent Platform versions
+. ../utils/config.sh
```
rspurgeon: My original thought when I suggested we have helper.sh source in config.sh, we just have all our scripts that users would run, just source in helper.sh.  That way it's just standard and the same across all scripts.  WDYT?

ybyzek: Given the move to config.env, I think current PR works.  Is that right or did I misunderstand your comment?

rspurgeon: i'm thinking this should be the following, just to standardize

```suggestion
. ../utils/helper.sh
```

andrewegel (approved):

microservices-orders/services-start.sh
rspurgeon: config.sh isn't a file anymore after the change to config.env

utils/config.env
```diff
@@ -0,0 +1,42 @@
...
+#   CP_VERSION_FULL=5.4.1
+#   GH_BRANCH=5.4.x
+#   JAR_VERSION=5.4.1-SNAPSHOT
+#   TAG=5.4.x-latest
+#   REPOSITORY=<account>.dkr.ecr.us-west-2.amazonaws.com/confluentinc   # confluentinc repository at the specified ECR registry, images must be pulled separately
+#####################################################
+
+CP_VERSION_MAJOR=5.4
+CP_VERSION_FULL=5.4.0
+GH_BRANCH=5.4.0-post
```
rspurgeon: _verrrry_ nitpicky, but FYC calling this `GIT_BRANCH` or `CP_BRANCH` as it doesn't really have much to do w/ github in particular... do not hold up the change for this unless you want

ybyzek: Ack and done with latest commit.

microservices-orders/services-start.sh
ybyzek: Great catch!

rspurgeon (approved):
LGTM

PR #418, opened 2020-01-22
DEVX-1315 sets pom.xml rest-utils-parent to 5.4.0 (1 of n )

rspurgeon:
This commit should be merged forward w/ strategy ours so downstream
branches do not receive this.

This commit will be followed shortly by a commit that updates an artifact
in the multi-datacenter demo to fix a naming issue, which we do
want to flow downstream.


ybyzek (approved):
Excellent catch @rspurgeon , LGTM

PR #419, opened 2020-01-22
DEVX-1315 multi-datacenter uses proper artifact for connect replicator jar

rspurgeon:


ybyzek (approved):
@rspurgeon verified it compiles, thanks for catching this!

PR #422, opened 2020-01-23
DEVX-1315: Further cleans up our pom.xml files

rspurgeon:


clients/avro/pom.xml
```diff
@@ -77,12 +79,15 @@ http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <build>
     <plugins>
       <plugin>
-      	<!-- Set the Java target version to 1.8 -->
+        <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-        <version>3.8.1</version>
+        <version>3.3</version>
```
ybyzek: @rspurgeon In some places `maven-compiler-plugin` version is 3.3 and in others it is 3.8.1.  Do we understand when to use which, and should we standardize?

rspurgeon: good eye.  I think I just made one of them the same as kafka-streams-examples and forgot to do the other ones the same.  I'm not aware of any material difference, but do agree it should be the same thorughout our repo.  I'll make it the latest (3.8.1) across the board unless you think it should be the same as ks-examples...

ybyzek: @rspurgeon I don't know the difference either to provide a recommendation for the latest or not. Can you circle with the engineering team to dig into this a bit more?  The reason is to be consistent with our engineering team in case we run into unexpected/unsupported issues.

ybyzek (approved):
LGTM thanks!

PR #423, opened 2020-01-24
Uses Kotlin collection functions in consumer

rspurgeon:
Also used Kotlin function for process exit as recommended by linter
Removed unused import


ybyzek:
I defer to @gAmUssA

gAmUssA (approved):
Lgtm tho for loop is quite ideomatic kotlin

PR #425, opened 2020-01-30
Docker image tags should be 5.5.x-latest in master

andrewegel:
Image tags in master should follow the `major.minor.x-latest` convention as opposed to the SNAPSHOT they currently are.

I also set the Readme image path in github to always point to master - theres no need to keep updating this value to anything else, even then you should really be hosing this image in a CDN, and serving it that way, rather than host binary files in a git repo.


mohnishbasha (approved):
lgtm

ybyzek (approved):
LGTM @andrewegel

PR #430, opened 2020-02-06
Fixed CP docs link for MRC

javabrett:
Existing link results in a recovery 302 to docs home - replaced with valid link.


ybyzek (approved):
@javabrett good catch, thanks for the PR

PR #432, opened 2020-02-11
DEVX-1463: Add REST Proxy to template delta configs for CCloud

ybyzek:


rspurgeon (approved):
lgtm, tested script directly locally

PR #437, opened 2020-02-17
DEVX-1512: Schema Registry Tutorial project compile fails on machine …

ybyzek:
…with Java 1.11

> 1.8 fails with maven-compiler-plugin version 3.8.1 (used in the SR tutorial and for some of the other examples), but works with maven-compiler-plugin version 3.3 (used in kafka-streams-examples)

Fix: change `java.version` from `1.8` to `8`

Note: I have validated that the four affected projects in this PR compile with java 8 and java 11


ghost (approved):
LGTM

rspurgeon (approved):
lgtm, built and ran tests for each as well as scoured for other poms (to no avail).

PR #440, opened 2020-02-24
DEVX-1192 Adds CONFLUENT prefix values to config.env

rspurgeon:


ybyzek (commented):
@rspurgeon Is this a trial PR or final change?  Should the last 4 lines of this file change?

andrewegel (approved):
I'll defer to your team's judgement about removing the last 4 lines. Otherwise LGTM

rspurgeon:
> @rspurgeon Is this a trial PR or final change? Should the last 4 lines of this file change?

My plan is to do this incrementally by having the new values put in by automation.  Nothing references them, and I'm not changing anything that references the existing names in this file, so there is no behavioral impact.   After this automation is proven out, we can further modify this file, scripts, etc... to take advantage of these new values.

ybyzek:
> My plan is to do this incrementally by having the new values put in by automation. Nothing references them, and I'm not changing anything that references the existing names in this file, so there is no behavioral impact. After this automation is proven out, we can further modify this file, scripts, etc... to take advantage of these new values.

Makes sense, thanks!

utils/config.env
```diff
@@ -1,6 +1,28 @@
...
+# The following values are provided by Confluent's
+# release automation process.  The values in these
+# names will be adopted in scripts and expressions
+# over time to construct other
+# values required by this repository
+#####################################################
+CONFLUENT=5.5.0-SNAPSHOT
+CONFLUENT_DOCKER_TAG=5.5.x-latest
+CONFLUENT_SHORT=5.5
+CONFLUENT_PREVIOUS=5.4.2
```
ybyzek: 1. Why is PREVIOUS necessary?
2. Is this by branch or date? e.g. Let's say 5.3.N is released after 5.4.N-1, then what would PREVIOUS be?

utils/config.env
```diff
@@ -1,6 +1,28 @@
...
+CONFLUENT_RELEASE_TAG_OR_BRANCH=5.5.x
+CONFLUENT_MAJOR=5
+CONFLUENT_MINOR=5
+CONFLUENT_PATCH=0
+KAFKA=5.5.0-SNAPSHOT
```
ybyzek, line 16: Are KAFKA and CE_KAFKA switched?

utils/config.env
rspurgeon: I'm not certain all of them are necessary, but what was discussed in the companion document was that we would accept all available values from the tools automation here and use what's required as we go.

utils/config.env
rspurgeon: I'm not certain... @andrewegel does it matter?  I pulled these from the companion spreadsheet.

utils/config.env
rspurgeon: @andrewegel can you answer @ybyzek 's question about PREVIOUS?

ybyzek: As we are pulling these values into a public-facing repo, we should have clarity on what they represent.  @andrewegel could you elaborate on the question regarding PREVIOUS?  Thanks!

utils/config.env
andrewegel: `KAFKA` is correct - `CE_KAFKA` would be `5.4.0-ce-SNAPSHOT` in this context.

utils/config.env
andrewegel: `CONFLUENT_PREVIOUS` is just the previous version - take it or leave it, its your choice.

andrewegel: As in this is the table:

`5.4.1` -> `5.4.0`
`5.4.0` -> `` (empty string)

utils/config.env
rspurgeon: I just hand typed this in so it will be replaced by an appropriate value after we merge this and automation kicks in, is that correct?  Does it need to be a valid value before that happens?

andrewegel: The value doesn't really matter, this is an idempotent process, if the script encounters any value, it'll just put in the right one.

utils/config.env
rspurgeon: Do you know what specifically defines the value?

andrewegel: https://github.com/confluentinc/release-tools/blob/master/confluent/release/versions.py#L692

ybyzek: @andrewegel thanks for the link.  Based on

```
    Return the previous patch version for the given version without any suffix, or None if this is the first
```

IIUC since there is no 5.5.0 yet, then this should be set to `None`

rspurgeon: I've updated the value here to empty string per @andrewegel comment above

ybyzek (approved):
> The value doesn't really matter, this is an idempotent process, if the script encounters any value, it'll just put in the right one.

Given this, PR approved.

PR #449, opened 2020-02-27
DEVX-1539 updates kafka go client import path

rspurgeon:
At some point the import path for the confluent-kafka-go client library has been updated.

### DIFF
--- a/clients/cloud/go/consumer.go
+++ b/clients/cloud/go/consumer.go
@@ -27,7 +27,7 @@ import (
 	"./ccloud"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


ybyzek:
cc: @confluentinc/clients for review

rigelbm:
I think the Go examples should be migrated to Go 11 Modules. You should use gopkg.in only if you don't have Modules configured. Here is how to do it: https://blog.golang.org/migrating-to-go-modules

rigelbm:
cc @edenhill

edenhill (approved):
LGTM.

edenhill:
Let's save modules instructions for later.

PR #453, opened 2020-03-03
DEVX-1561 adds config.env values to be replaced by pattern matching in tools

rspurgeon:
This upgrades the file to have the values provided by automation after https://github.com/confluentinc/common-tools/pull/362


utils/config.env
```diff
@@ -1,6 +1,28 @@
...
+# The following values are provided by Confluent's
+# release automation process.  The values in these
+# names will be adopted in scripts and expressions
+# over time to construct other
+# values required by this repository
+#####################################################
+CONFLUENT=5.4.1-SNAPSHOT
+CONFLUENT_DOCKER_TAG=5.4.x-latest
+CONFLUENT_SHORT=5.4
+CONFLUENT_PREVIOUS=5.3.0
```
ybyzek: Not a big deal but IIUC this should be `5.4.0`

ybyzek (approved):
Minor comment and LGTM.  Thank you!

utils/config.env
andrewegel: It will actually be an empty string if its a version that ends in a `.0`

rspurgeon: So then what @ybyzek says is correct... if this is 5.4.1 then PREVIOUS is 5.4.0?

andrewegel: Yes

PR #468, opened 2020-03-17
no need to expose 29092

rmoff:
No need to expose 29092, since it's only going to work within the docker network. For external connections use 9092 (already exposed).
Partially a nit, but also makes it more ambiguous for people trying to understand it.

### DIFF
--- a/cp-all-in-one-community/docker-compose.yml
+++ b/cp-all-in-one-community/docker-compose.yml
@@ -18,7 +18,6 @@ services:
     depends_on:
       - zookeeper
     ports:
-      - "29092:29092"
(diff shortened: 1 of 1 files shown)


ybyzek (changes requested):
@rmoff this docker-compose.yml file is used by the CP quickstart at https://docs.confluent.io/current/quickstart/cos-docker-quickstart.html . Can you please do a sanity check on the quickstart to ensure it still works?

ybyzek:
Closing jira because cp-all-in-one has moved to https://github.com/confluentinc/cp-all-in-one

PR #482, opened 2020-03-20
DEVX-1623: Add C3, convert kafkacat to confluent CLI

ybyzek:


bbejeck:
Changes LGTM, but I'll do another review once DEVX-1625 is resolved.

PR #484, opened 2020-03-23
DEVX-1625: Revert window start for clickstream

bbejeck:
Looks like changes introduced in 5.1.0 "WindowStart() as..." are the culprit as the window start timestamp never advances after the start


ybyzek (approved):
💯 validated that this works, thanks @bbejeck .

PR #486, opened 2020-03-25
DEVX-1565 Updates config.env to use tool provided values

rspurgeon:
The existing values in this file are used by various scripts and tools (docker),
so as a first step I made our previous variables, like CP_VERSION_MAJOR, derive
its value from the CONFLUENT_* values which are managed by automation.  This allows
for scripts, like ./music/start.sh:check_running_cp ${CP_VERSION_MAJOR}, to work
w/ the automated values instead of a manually managed one.

### DIFF
--- a/utils/config.env
+++ b/utils/config.env
@@ -41,8 +41,8 @@ KAFKA_BRANCH=2.4
 #   REPOSITORY=<account>.dkr.ecr.us-west-2.amazonaws.com/confluentinc   # confluentinc repository at the specified ECR registry, images must be pulled separately
 #####################################################
 
-CP_VERSION_MAJOR="$CONFLUENT_MAJOR.$CONFLUENT_MINOR"
-CP_VERSION_FULL="$CONFLUENT_MAJOR.$CONFLUENT_MINOR.$CONFLUENT_PATH"
+CP_VERSION_MAJOR="$CONFLUENT_SHORT"
+CP_VERSION_FULL="$CONFLUENT_MAJOR.$CONFLUENT_MINOR.$CONFLUENT_PATCH"
 GIT_BRANCH=$CONFLUENT_RELEASE_TAG_OR_BRANCH
 JAR_VERSION=$CONFLUENT
 
--- a/utils/config.env
+++ b/utils/config.env
@@ -41,7 +41,7 @@ KAFKA_BRANCH=2.4
 #   REPOSITORY=<account>.dkr.ecr.us-west-2.amazonaws.com/confluentinc   # confluentinc repository at the specified ECR registry, images must be pulled separately
 #####################################################
 
-CP_VERSION_MAJOR="$CONFLUENT_SHORT"
+CP_VERSION_MAJOR=$CONFLUENT_SHORT
 CP_VERSION_FULL="$CONFLUENT_MAJOR.$CONFLUENT_MINOR.$CONFLUENT_PATCH"
 GIT_BRANCH=$CONFLUENT_RELEASE_TAG_OR_BRANCH
 JAR_VERSION=$CONFLUENT
--- a/utils/config.env
+++ b/utils/config.env
@@ -41,10 +41,10 @@ KAFKA_BRANCH=2.4
 #   REPOSITORY=<account>.dkr.ecr.us-west-2.amazonaws.com/confluentinc   # confluentinc repository at the specified ECR registry, images must be pulled separately
 #####################################################
 
-CP_VERSION_MAJOR=5.5
-CP_VERSION_FULL=5.5.0
-GIT_BRANCH=master
-JAR_VERSION=5.5.0-SNAPSHOT
+CP_VERSION_MAJOR="$CONFLUENT_MAJOR.$CONFLUENT_MINOR"
+CP_VERSION_FULL="$CONFLUENT_MAJOR.$CONFLUENT_MINOR.$CONFLUENT_PATH"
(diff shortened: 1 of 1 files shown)


utils/config.env
```diff
@@ -41,10 +41,10 @@ KAFKA_BRANCH=2.4
 #   REPOSITORY=<account>.dkr.ecr.us-west-2.amazonaws.com/confluentinc   # confluentinc repository at the specified ECR registry, images must be pulled separately
 #####################################################
 
-CP_VERSION_MAJOR=5.5
-CP_VERSION_FULL=5.5.0
-GIT_BRANCH=master
-JAR_VERSION=5.5.0-SNAPSHOT
+CP_VERSION_MAJOR="$CONFLUENT_MAJOR.$CONFLUENT_MINOR"
```
ybyzek: Why is `CP_VERSION_MAJOR` necessary if there is already `CONFLUENT_SHORT`?

If `CP_VERSION_MAJOR` is used elsewhere in the repo, it should be replaced with `CONFLUENT_SHORT`

utils/config.env
```diff
@@ -41,10 +41,10 @@ KAFKA_BRANCH=2.4
...
+CP_VERSION_FULL="$CONFLUENT_MAJOR.$CONFLUENT_MINOR.$CONFLUENT_PATH"
```
ybyzek: Is this supposed to be `CONFLUENT_PATCH`?

Can you print these all out to verify they have meaningful values (ie no other spelling issues)?

utils/config.env
```diff
@@ -41,10 +41,10 @@ KAFKA_BRANCH=2.4
...
+GIT_BRANCH=$CONFLUENT_RELEASE_TAG_OR_BRANCH
```
ybyzek, line 46: Why is `GIT_BRANCH` necessary if there is already `CONFLUENT_RELEASE_TAG_OR_BRANCH`?

If `GIT_BRANCH` is used elsewhere in the repo, it should be replaced with `CONFLUENT_RELEASE_TAG_OR_BRANCH`.

utils/config.env
```diff
@@ -41,10 +41,10 @@ KAFKA_BRANCH=2.4
...
+JAR_VERSION=$CONFLUENT
```
ybyzek, line 47: Same comment as above.  This variable was created before the discussion with the automation tools and should be removed.

utils/config.env
```diff
@@ -57,7 +57,7 @@ JAR_VERSION=5.5.0-SNAPSHOT
 
 # TAG - image tag.
 # The ':' which separates the image name from the TAG is not required here
-TAG=5.5.x-latest
+TAG=$CONFLUENT_DOCKER_TAG
```
ybyzek, line 60: Same comment as above.  This variable was created before the discussion with the automation tools and should be removed.

ybyzek (changes requested):
Let's remove redundancy in variables

utils/config.env
rspurgeon: I ran demo tests so i'm not sure how this got by unless I inserted a typo prior to committing somehow

utils/config.env
rspurgeon: I really prefer to make minimum possible changes per commit / PR.   The associated JIRA is also very specific and not related to the projects (music & microservices) that depend on GIT_BRANCH and JAR_VERSION, so I would prefer to leave those to other commits.   Small PRs are easier to find issues in, like you did above.

utils/config.env
rspurgeon: Same as below, in an effort to keep the changes minimal in this PR, I just wanted to make the demos work w/ the values presented to us and this is the minimal change I could think of.  If we want to remove redundant names, I suggest we remove them in another PR to keep changes small.

rspurgeon: I addressed to use the SHORT version of the provided value, nice catch.  thanks

utils/config.env
rspurgeon: I've added DEVX-1641 to track this

rspurgeon:
```
source ./utils/config.env; set | grep 'JAR_VERSION\|GIT_BRANCH\|CONFLUENT\|CP_\|KAFKA\|TAG'
CE_KAFKA=2.4
CONFLUENT=5.5.0-SNAPSHOT
CONFLUENT_DOCKER_TAG=5.5.x-latest
CONFLUENT_HOME=/Users/rspurgeon/software/cp/confluent-5.5.0-SNAPSHOT/
CONFLUENT_MAJOR=5
CONFLUENT_MINOR=5
CONFLUENT_PATCH=0
CONFLUENT_PREVIOUS=''
CONFLUENT_RELEASE_TAG_OR_BRANCH=5.5.x
CONFLUENT_SHORT=5.5
CP_VERSION_FULL=5.5.0
CP_VERSION_MAJOR=5.5
GIT_BRANCH=5.5.x
JAR_VERSION=5.5.0-SNAPSHOT
KAFKA=5.5.0-SNAPSHOT
KAFKA_BRANCH=2.4
KAFKA_RELEASE_TAG_OR_BRANCH=2.4
KAFKA_SHORT=2.4
TAG=5.5.x-latest
```

utils/config.env
rspurgeon: printed out the calculated values and added a comment below

ybyzek (approved):
Ack LGTM, thanks @rspurgeon

PR #489, opened 2020-03-26
DEVX-1192 updates music demo docker-compose.yaml to leverage config.env

rspurgeon:
This commit does not resolve entirely DEVX-1192, just for music alone and only relates to the docker changes.  More commits will follow for other demos.

### DIFF
--- a/music/.env
+++ b/music/.env
@@ -1,0 +1,1 @@
+../utils/config.env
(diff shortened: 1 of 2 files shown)


ybyzek (approved):
LGTM, thanks @rspurgeon

PR #494, opened 2020-03-26
DEVX-1192 updates postgres debezium demo docker-compose.yaml to lever…

rspurgeon:
…age config.env


ybyzek (approved):
LGTM

PR #495, opened 2020-03-26
DEVX-1192 updates replicator security source ssl docker-compose.yaml …

rspurgeon:
…to leverage config.env


ybyzek (approved):
LGTM thanks!

PR #498, opened 2020-03-26
DEVX-1192 Updates the cp-all-in-one docker-compose files to use config.env

rspurgeon:


ybyzek (approved):
LGTM

rmoff:
@rspurgeon is there background to this change? I was about to link to this from a StackOverflow answer, but am concerned that it'll be more confusing since it's not a standalone example, since it now depends on `.env` and thus also `../utils/config.env` etc. For a user already struggling with Kafka and Docker, this may be a step too far ;-) 

If there's good reasons to leave this as-is then maybe we should also publish standalone Docker Compose examples, WDYT?

rspurgeon:
@rmoff The associated Jira will have more details, but generally these kinds of changes helps us with a growing maintenence problem in addition to testing both publically available and internal only build artifacts (historical and 'future' versions).  Our team has multiple repos w/ many version numbers and other variable bits like docker registries.  The number of product versions is also growing as products are bifurcating from the Confluent Platform.  So this change was part of a broader effort to lessen the maintenence cost of supporting that matrix.  We also debated internally on our team the tradeoff between maintence and readability but landed on the need to reduce our maintence issues.

The docker-compose system still maintains "standalone" behavior as linking in an enviornment file is supported and documented by docker-compose itself.    The "out of the box" behavior for users of "docker-compose up" should still work, but I get your concern about readability.

What if we added comments in the docker-compose files to explain how the linkage works and where to find the values?   tag @ybyzek

PR #499, opened 2020-03-27
DEVX-1192 updates multidatacenter docker-compose.yaml to leverage con…

rspurgeon:
…fig.env


ybyzek (approved):
LGTM

PR #507, opened 2020-03-27
DEVX-1641 docker-compose.yml tags read from tools provided values

rspurgeon:
The redirection previously in the .env files does not work with docker compose, this fixes that
and works towards removing redundant names across the repo.  This commit does not
resolve the JIRA entirely.


ybyzek (approved):
LGTM.

PR #508, opened 2020-03-27
DEVX-1641 updates start.sh scripts that validate CP version to use tool value

rspurgeon:
instead of the prvious redirected one, which was redundant.


ybyzek (approved):
LGTM

PR #511, opened 2020-03-27
DEVX-1650: Consolidate get-kafka-streams-examples.sh into library

ybyzek:


rspurgeon (approved):
LGTM.  Unfortunately this still takes a bit of time, but IMO the improvements to the build time need to be made in the dependent project instead of trying to improve the time here.

PR #512, opened 2020-03-28
DEVX-1652: Replace deprecated way to name ksqlDB cluster in C3 config…

ybyzek:
… in cp-all-in-one


music/docker-compose.yml
```diff
@@ -159,8 +159,8 @@ services:
     environment:
       CONTROL_CENTER_BOOTSTRAP_SERVERS: 'kafka:29092'
       CONTROL_CENTER_ZOOKEEPER_CONNECT: 'zookeeper:32181'
-      CONTROL_CENTER_KSQL_URL: "http://ksql-server:8088"
-      CONTROL_CENTER_KSQL_ADVERTISED_URL: "http://localhost:8088"
+      CONTROL_CENTER_KSQL_MUSIC_URL: "http://ksql-server:8088"
```
rspurgeon, line 162: FYI For music, the `confluent local` version of the demo doesn't match the ksqlDB rename, if we care.

rspurgeon (approved):
LGTM minor comment on the music demo

music/docker-compose.yml
ybyzek: @rspurgeon it doesn't match and it won't because it shouldn't modify that properties file.  However there's still a bug there (MMA-6998) that we can follow-up on separately

PR #513, opened 2020-03-28
DEVX-1540: Python client example cannot be run OOTB to CCloud without ssl.ca.location

ybyzek:


rspurgeon (approved):
LGTM, tested locally w/ much success :-)

PR #516, opened 2020-03-28
DEVX-1653: ksql-datagen: replace maxInterval with msgRate

ybyzek:

PR #517, opened 2020-03-30
Switch DataGen  `maxInterval` to `msgRate`

big-andy-coates:
Update use of DataGen to remove use of deprecated `maxInterval` and replace with new `msgRate` command line arg.

* the old `maxInterval` introduced a sleep between each message produced.
* the new `msgRate` throttles the message production to the required message rate per second.

Note: the `DatagenConnector` connector still looks to use `maxInterval` and so its config hasn't been updated.

Note: I've not run these examples to test this change.

Note: I also noticed that this issue referenced a lot: https://github.com/confluentinc/ksql/issues/1031, which necessitated listing all the columns in CREATE STREAM/TABLE commands, but that issue is resolved, so may be worth cleaning those up...?

### DIFF
--- a/clients/cloud/ksql-datagen/docker-compose.yml
+++ b/clients/cloud/ksql-datagen/docker-compose.yml
@@ -7,7 +7,7 @@ services:
     container_name: ksql-datagen
     volumes:
       - $PWD/delta_configs/ksql-datagen.delta:/tmp/ksql-datagen.delta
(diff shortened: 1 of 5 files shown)


ybyzek:
@big-andy-coates is this a duplicate of https://github.com/confluentinc/examples/pull/516 ?

big-andy-coates:
Yep - sorry, not spotted that. Closing...

However, note my comment about https://github.com/confluentinc/ksql/issues/1031

PR #519, opened 2020-03-30
Update microservices-orders inline with latest ksqlDB syntax

big-andy-coates:
Note: I've ensured the ksql commands are valid, but I have not run through the entire example.

Part of work for https://github.com/confluentinc/ksql/issues/4144

### DIFF
--- a/microservices-orders/ksql.commands
+++ b/microservices-orders/ksql.commands
@@ -9,10 +9,10 @@ CREATE STREAM customers_with_wrong_format_key (id bigint, firstname varchar, las
 --2. Derive a new stream with the required key changes.
 --The CAST statement converts the key to the required format.
 --The PARTITION BY clause re-partitions the stream based on the new, converted key.
-CREATE STREAM customers_with_proper_key WITH (KAFKA_TOPIC='customers-with-proper-key', PARTITIONS=1) AS SELECT CAST(id as VARCHAR) as customerid, firstname, lastname, email, address, level FROM customers_with_wrong_format_key PARTITION BY customerid;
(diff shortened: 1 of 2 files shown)


microservices-orders/ksql.commands
```diff
@@ -9,10 +9,10 @@ CREATE STREAM customers_with_wrong_format_key (id bigint, firstname varchar, las
 --2. Derive a new stream with the required key changes.
 --The CAST statement converts the key to the required format.
 --The PARTITION BY clause re-partitions the stream based on the new, converted key.
-CREATE STREAM customers_with_proper_key WITH (KAFKA_TOPIC='customers-with-proper-key', PARTITIONS=1) AS SELECT CAST(id as VARCHAR) as customerid, firstname, lastname, email, address, level FROM customers_with_wrong_format_key PARTITION BY customerid;
+CREATE STREAM customers_with_proper_key WITH (KAFKA_TOPIC='customers-with-proper-key', PARTITIONS=1) AS SELECT CAST(id as BIGINT) as customerid, firstname, lastname, email, address, level FROM customers_with_wrong_format_key PARTITION BY CAST(id as BIGINT);
```
big-andy-coates, line 12: `PARTITION BY` is now applied in a consistent manner to `GROUP BY`, i.e. its applied to the source _before_ the projection.   `PARTITION BY` can also now be an expression.

Note: the old version was a bit weird as it was casting a varchar to a varchar! It now casts it to a `BIGINT`, which matches the id type in the other source.

ybyzek (approved):
LGTM thank you!

PR #521, opened 2020-03-30
Update mysql-debezium inline with latest ksqlDB syntax

big-andy-coates:
Note: I've only tested the sql statements, not the entire example.

Part of work for https://github.com/confluentinc/ksql/issues/4144


ybyzek (approved):
LGTM thanks @big-andy-coates

PR #524, opened 2020-03-31
DEVX-1631 and DEVX-1486: cp-ksql-server to cp-ksqldb-server; fix remo…

ybyzek:
…ve KSQL_LOG4J_OPTS


cp-all-in-one-cloud/docker-compose.yml
```diff
@@ -18,15 +18,14 @@ services:
       SCHEMA_REGISTRY_KAFKASTORE_SASL_MECHANISM: "PLAIN"
 
   ksql-server:
```
rspurgeon: I guess we should we go ahead and rename the docker-compose services as well, either here or in a subsequent PR.

ybyzek: @rspurgeon it makes sense to do it in this PR.  I'll take care of it.

rspurgeon: @ybyzek bear in mind if we have documentation around that informs users to 
>docker exec -it ksql-cli ksql http://ksql-server:8088

this will break w/ the ksqdb-server change suggested above.

ybyzek: Ack and done https://github.com/confluentinc/examples/pull/524/commits/157407beae657fa39575fe0e00262144fe087e44

PR #526, opened 2020-03-31
DEVX-1660: Implement common retry function for deterministic wait for…

ybyzek:
… services

### DIFF
--- a/utils/helper.sh
+++ b/utils/helper.sh
@@ -723,10 +723,10 @@ check_control_center_up() {
 
 check_topic_exists() {
   containerName=$1
-  zkConn=$2
+  brokerConn=$2
   topic=$3
 
-  docker-compose exec $containerName kafka-topics --zookeeper $zkConn --describe --topic $topic >/dev/null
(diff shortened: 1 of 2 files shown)


ybyzek:
@rspurgeon given the timing issues you reported yesterday, I hope this resolves it in a deterministic way

utils/helper.sh
```diff
@@ -679,6 +679,68 @@ function wait_for_connector_up() {
...
+  fi
+  return 0
+}
+
+check_topic_exists() {
+  containerName=$1
+  zkConn=$2
+  topic=$3
+
+  docker-compose exec $containerName kafka-topics --zookeeper $zkConn --describe --topic $topic >/dev/null
```
rspurgeon: I think we should be use the `--bootstrap-server` version of this command

`kafka-topics --bootstrap-server broker-dc1:9091 --describe --topic topic1` for example

ybyzek: @rspurgeon 💯 and fixed

rspurgeon (approved):
LGTM

PR #530, opened 2020-04-01
DEVX-1667: Dedupe pipeline demo-scene demos from confluentinc/examples

ybyzek:
* https://github.com/confluentinc/examples/blob/DEVX-1667/README.md#additional-demos
* Deleted entire folders/demos in favor of pointing users to demo-scene directly


rspurgeon (approved):
LGTM ❗️

PR #533, opened 2020-04-01
DEVX-1641 cleans up utils/config.env to match new cp-demo file

rspurgeon:
This removes some outdated & unecessary comments from the config.env file and makes it match what has been added to the cp-demo project.


ybyzek (approved):
LGTM

PR #534, opened 2020-04-01
DEVX-1316 applies minor tidy to helper.sh

rspurgeon:
The bash syntax `command || { or commands }` is a little more direct
versus the `command; if [ $? != 0 ]; then { or commands }`.

This ties the failure of the command directly to the resulting action
and helps prevents someone from accidentally doing something similar inserting a line between:

```
gcloud auth activate-service-account --key-file $GCS_CREDENTIALS_FILE
echo "$GC_CREDENTIALS_FILE for debugging purposes"
if [[ "$?" -ne 0]]; then
```

which would result in returning the retcode of echo accidentally.

### DIFF
--- a/utils/helper.sh
+++ b/utils/helper.sh
@@ -105,7 +105,7 @@ function check_ccloud_logged_in() {
 }
 
 function version_gt() { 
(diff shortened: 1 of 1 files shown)


ybyzek (approved):
LGTM

PR #535, opened 2020-04-02
DEVX-987: Update cp-all-in-one-cloud to notify users they need to man…

ybyzek:
…ually create the config properties file


rspurgeon (approved):
lgtm

PR #541, opened 2020-04-06
DEVX-1683: tune configuration for ksqlDB logging and embedded Connect…

ybyzek:
… worker


bbejeck (approved):
Thanks @ybyzek LGTM

microservices-orders/docker-compose.yml
```diff
@@ -125,7 +125,6 @@ services:
       - "8088:8088"
     environment:
       KSQL_CONFIG_DIR: "/etc/ksql"
-      KSQL_LOG4J_OPTS: "-Dlog4j.configuration=file:/etc/ksql/log4j-rolling.properties"
```
bbejeck, line 128: One quick question - why not update to `KSQL_LOG4J_OPTS: "-Dlog4j.configuration=file:/etc/ksql/log4j.properties"` vs. removing? Unless that is the default setting.  I wouldn't block the PR for this though.

agavra (approved):

microservices-orders/docker-compose.yml
ybyzek: @bbejeck my thinking was that it's an optional config that adds to "noise".  Logging works without it now (given a recent fix from @agavra ).

PR #548, opened 2020-04-08
DEVX-1692: Add Docker-based cp-all-in-one to cp-quickstart

ybyzek:


cp-quickstart/stop-docker.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# Source library 
+. ../utils/helper.sh
+
+docker-compose down
```
rspurgeon: I think we want to add `-v` to cleanup volumes (present or future) used by this project.

.gitignore
```diff
@@ -20,3 +20,4 @@ log.*
 *.csr
 *.srl
 *_creds
+cp-quickstart/docker-compose.yml
```
rspurgeon: FYC: It might be easier across our many sub-projects to keep `.gitignore` files in subfolders themselves instead of centrally at the root.   Also might be better / easier if we end up moving / or deprecating projects if they have their own .gitignore.

rspurgeon (approved):
LGTM w/ only two minor comments.  I only think the `docker-compose down -v` change is necessary unless you want to take the other

cp-quickstart/stop-docker.sh
ybyzek: Ack and done

.gitignore
ybyzek: Great idea @rspurgeon .  Done

PR #549, opened 2020-04-08
DEVX-1693: Add Docker-based cp-all-in-one-community to cp-quickstart

ybyzek:


cp-quickstart/start-docker-community.sh
```diff
@@ -0,0 +1,35 @@
...
+docker-compose up -d
+
+# Verify Kafka Connect worker has started
+MAX_WAIT=120
+echo "Waiting up to $MAX_WAIT seconds for Connect to start"
+retry $MAX_WAIT check_connect_up connect || exit 1
+sleep 2 # give connect an exta moment to fully mature
+echo "connect has started!"
+
+wget https://github.com/confluentinc/kafka-connect-datagen/raw/master/config/connector_pageviews_cos.config
```
rspurgeon: should we make this consistent with the wget above in use of the `--output-file / -O` flag and the use of the githubusercontent.com url for downloading raw content?

ybyzek: I added `-O`.

However, the rest of the command is consistent with what the CP quickstart does (see https://docs.confluent.io/current/quickstart/cos-docker-quickstart.html#step-3-install-a-ak-connector-and-generate-sample-data), so we should leave it as-is in this PR and not change to `raw.githubusercontent.com`

rspurgeon (approved):
lgtm

PR #551, opened 2020-04-09
DEVX-1643 ccloud/beginner-cloud uses KAFKA_CONNECT_DATAGEN variables

rspurgeon:


rspurgeon:
@ybyzek can you please verify what I'm seeing in that the Dockerfile I deleted from this isn't needed?

ybyzek (approved):
LGTM.  Agree `Dockerfile` can be deleted (I believe it's a remnant from previous iteration of this demo).

PR #558, opened 2020-04-09
DEVX-1700: Python client example: fail fast if cannot create topic

ybyzek:


rspurgeon:
@ybyzek I posted downstream PR #559 to here which addresses the concern about an invalid message to the user on production errors.  Once that's merged into here I'm good w/ this change.

rspurgeon (approved):
LGTM

PR #566, opened 2020-04-16
DEVX-1496 ccloud docker demo adds ACLs

rspurgeon:
and minor UX improvements.  Also uses -v in docker-compose down
instead of removing the users other dangling volumes.

### DIFF
--- a/ccloud/start-docker.sh
+++ b/ccloud/start-docker.sh
@@ -25,7 +25,11 @@ export CONFIG_FILE=~/.ccloud/config
 check_ccloud_config $CONFIG_FILE || exit 1
 check_ccloud_version 0.264.0 || exit 1
 check_ccloud_logged_in || exit 1
-printf "Done\n"
+printf "Done\n\n"
+
+echo ====== Cleaning up previous run
+./stop-docker.sh
+printf "\nDone with cleanup\n\n"
 
 echo ====== Generate CCloud configurations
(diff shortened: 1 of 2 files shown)


ccloud/start-docker.sh
```diff
@@ -24,40 +25,59 @@ export CONFIG_FILE=~/.ccloud/config
...
+echo ====== Validate credentials to Confluent Cloud Schema Registry
 validate_confluent_cloud_schema_registry $SCHEMA_REGISTRY_BASIC_AUTH_USER_INFO $SCHEMA_REGISTRY_URL || exit 1
+printf "Done\n\n"
 
+echo ====== Starting local services in Docker
 docker-compose up -d
+printf "\n"
 
 echo "Sleeping 120 seconds to wait for all services to come up"
+printf "\n"
```
ybyzek: Instead of the `sleep 120` (below), can you implement the equivalent of https://github.com/confluentinc/examples/blob/5.5.x/ccloud/start.sh#L117-L120

The main difference being the wait time and use this function instead https://github.com/confluentinc/examples/blob/5.5.x/utils/helper.sh#L708

rspurgeon: Nice, for sure...good eye.  WIP

rspurgeon: @ybyzek please see revision

ybyzek (approved):
LGTM thanks @rspurgeon .  So glad this example works for service accounts now :)

PR #567, opened 2020-04-16
fixed Makefile-impl USER variable issue on some Cloud Shell Envrionments

jasonsmithio:
In some Cloud Shell setups, using $USER causes an error. I replaced the variable

### DIFF
--- a/kubernetes/gke-base/Makefile-impl
+++ b/kubernetes/gke-base/Makefile-impl
@@ -6,7 +6,7 @@ include $(GKE_BASE_MKFILE_DIR)../common/Makefile
 
 GCP_PROJECT_ID ?= $(shell gcloud config list --format 'value(core.project)')
 
-GKE_BASE_CLUSTER_ID ?= cp-examples-operator-$(USER)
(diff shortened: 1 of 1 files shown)


rspurgeon:
Thanks for the PR @TheJaySmith!  I'm curious which cloud shell environment you ran this on to surface the issue?

jasonsmithio:
qwiklabs.com environment. We aren't sure why yet

rspurgeon:
> qwiklabs.com environment. We aren't sure why yet

Thanks for the reply.  One thing I like about the usage of USER is that we may have multiple simultaneous runners of our demo targeting the same GCP project.   Any thoughts on a different way to distinquish demo runners that would work for your environment?

jasonsmithio:
One issue is the character limit in GKE for cluster names. Our username for QwikLabs are long so tagging that onto an already lengthy cluster name exceeds the limit. Perhaps we can allow people to add their own tags to the cluster name?

rspurgeon:
@TheJaySmith I'm going to close this for now, if we get another instance of the use of $USER being an issue I will revist an alternative approach.  Thanks

PR #569, opened 2020-04-17
Quick cleanup in gke-base steps to properly cleanup zk deployments

rspurgeon:
ZK deployments were not cleaning up properly in k8s demos.  Added back in some explicit commands to clean them up on demo destruction.


ybyzek (approved):
Not validated but LGTM

PR #580, opened 2020-04-20
DEVX-1730: Streams app development tutorial, mircroservices app fails…

ybyzek:
… to start up

### DIFF
--- a/microservices-orders/docker-compose.yml
+++ b/microservices-orders/docker-compose.yml
@@ -169,6 +169,7 @@ services:
       - "18894:18894"
     volumes:
       - $PWD/services-start.sh:/tmp/services-start.sh
+      - $PWD/config.env:/tmp/config.env
     command: "bash -c 'echo Waiting for Kafka to be ready... && \
                        cub kafka-ready -b broker:9092 1 30 && \
                        echo Waiting for Confluent Schema Registry to be ready... && \
--- a/microservices-orders/docker-compose.yml
+++ b/microservices-orders/docker-compose.yml
@@ -169,7 +169,7 @@ services:
       - "18894:18894"
     volumes:
       - $PWD/services-start.sh:/tmp/services-start.sh
-      - $PWD/config.env:/tmp/config.env
+      - $PWD/.env:/tmp/config.env
     command: "bash -c 'echo Waiting for Kafka to be ready... && \
                        cub kafka-ready -b broker:9092 1 30 && \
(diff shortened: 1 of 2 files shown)


microservices-orders/docker-compose.yml
```diff
@@ -169,6 +169,7 @@ services:
       - "18894:18894"
     volumes:
       - $PWD/services-start.sh:/tmp/services-start.sh
+      - $PWD/config.env:/tmp/config.env
```
rspurgeon: There is no `config.env` file in $PWD, maybe you meant $PWD/../utils/config.env or similar?  The result of this testing the current way is that /tmp/config.env is an empty directory

```
streams@microservices-app:/build$ ls -la /tmp
total 16
drwxrwxrwt 1 root    root    4096 Apr 21 13:03 .
drwxr-xr-x 1 root    root    4096 Apr 21 13:03 ..
drwxr-xr-x 2 streams streams   64 Apr 21 13:03 config.env
drwxr-xr-x 2 streams streams 4096 Apr 21 13:04 hsperfdata_streams
-rwxr-xr-x 1 streams streams  893 Apr 21 12:52 services-start.sh

```

ybyzek: @rspurgeon the script (see below) sources `/tmp/config.env` so `$PWD` doesn't come into the picture (on the container itself)

rspurgeon: $PWD is the volume map on the host, but there is no config.env file on the host there, so the result is inside the container there is no config.env file at /tmp.  Docker makes an assumption that config.env is a directory if there is no file on the host at the volume point.  So the result inside the container is a directory, not a file, at /tmp/config.env.  As a result, the source command inside services-start.sh can't be sourcing anything meaningful.  Here is another command showing the issue from inside the container. 

```
streams@microservices-app:/build$ cat /tmp/config.env
cat: /tmp/config.env: Is a directory
```

ybyzek: @rspurgeon thanks for surfacing this.  Source of confusion was an unchecked file on my local host.  Can you please verify PR with latest commit?

rspurgeon: 👍 
```
docker-compose exec microservices-app bash
streams@microservices-app:/build$ cat /tmp/config.env
#####################################################
# Values in this file will be used to replace variables in script files
...
```

rspurgeon (approved):
LGTM

PR #599, opened 2020-04-28
DEVX-1344: Set TIMESTAMPS_TOPIC_REPLICATION_FACTOR_CONFIG=1 and other cleanup in ccloud demo

ybyzek:
`__consumer_timestamps` is in the origin cluster


utils/helper.sh
```diff
@@ -839,16 +839,8 @@ function create_connect_topics_and_acls() {
...
-  for topic in __consumer_timestamps ; do
-    ccloud kafka topic create $topic &>/dev/null
-    ccloud kafka acl create --allow --service-account $serviceAccount --operation WRITE --topic $topic --prefix
-    ccloud kafka acl create --allow --service-account $serviceAccount --operation READ --topic $topic --prefix
-  done
   ccloud kafka acl create --allow --service-account $serviceAccount --operation READ --consumer-group connect-replicator
   ccloud kafka acl create --allow --service-account $serviceAccount --operation describe --cluster-scope
-  ccloud kafka acl create --allow --service-account $serviceAccount --operation describe --topic pageviews
-  ccloud kafka acl create --allow --service-account $serviceAccount --operation describe-configs --topic pageviews
-  ccloud kafka acl create --allow --service-account $serviceAccount --operation describe --topic __consumer_timestamps
```
rspurgeon, line 857: +1 for moving the demo specific topic configurations out of the shared function into the demo itself.  I'm curious why you're also removing the describe acl on __consumer_timestamps.  Has something changed to make that no longer necessary?

ybyzek: `__consumer_timestamps` was never required in the first place.  This topic is created on the origin cluster, not destination cluster.

rspurgeon (approved):
LGTM

PR #601, opened 2020-04-28
DEVX-1658: Don't blanket remove docker volumes from user's systems

ybyzek:


multiregion/README.md
```diff
@@ -449,6 +449,7 @@ Stop the demo and all Docker containers.
 # Clean up the Docker environment
 for c in $(docker container ls -q --filter "name=pumba"); do docker container stop "$c" && docker container rm "$c"; done
 docker-compose down -v --remove-orphans
+# More aggressive cleanup
```
rspurgeon, line 452: FYC: It might be better to recommend the docker provided prune commands, like:
`docker volume prune`
At least it gives them a proper warning and prompt before removing data volumes.

rspurgeon:
LGTM w/ one comment FYC

rspurgeon (approved):
LGTM w/ one comment FYC

multiregion/README.md
ybyzek: Agree and done in latest commit.

PR #602, opened 2020-04-28
DEVX-1764: Add Docker file for Python client

ybyzek:


clients/cloud/python/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu:18.04
+
+RUN apt-get update && \
+    apt-get install -y python3 python3-pip curl vim
+RUN pip3 install -U confluent-kafka
+RUN pip3 install -U confluent-kafka[avro]
```
russau: Minor: is it more "pythonic" to include a `requirements.txt` or [pipfile](https://github.com/pypa/pipfile).

ybyzek: Thanks @russau .   @javabrett was making a similar point in Slack.  Please see latest commit that moves to` requirements.txt`

rspurgeon:
@ybyzek FYC I was able to reduce the final image size by ~ 300MB and significantly improve the build time by changing the base image to `python:3-slim`.  This removes the need to use apt-get to install python and pip and update alternatives, they are baked into the official base images.  The other required change is then to add `bash` to your sample command because the python image uses python as the entrypoint.  I think this results in a more streamlined demo.

Here is the Dockerfile I tested
```
FROM python:3-slim

COPY requirements.txt /tmp/requirements.txt
RUN pip3 install -U -r /tmp/requirements.txt

COPY *.py ./
```

`cloud-demo-python                                                                        latest                        78894d19557f        5 minutes ago       247MB
`
vs
`cloud-demo-python                                                                        latest                        4c6cbcdb3d27        24 seconds ago      566MB`

The run command:
`docker run -v $PWD/librdkafka.config:/root/.confluent/librdkafka.config -it --rm cloud-demo-python bash`

Tested successfully:
```
root@d35ce8e17470:/# ./producer.py -f $HOME/.confluent/librdkafka.config -t test1
./producer.py:70: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
  p.produce(topic, key=record_key, value=record_value, on_delivery=acked)
Producing record: alice {"count": 1}
Producing record: alice {"count": 2}
Producing record: alice {"count": 3}
Producing record: alice {"count": 4}
Producing record: alice {"count": 5}
Producing record: alice {"count": 6}
Producing record: alice {"count": 7}
Producing record: alice {"count": 8}
Producing record: alice {"count": 9}
Produced record to topic test1 partition [0] @ offset 10
Produced record to topic test1 partition [0] @ offset 11
Produced record to topic test1 partition [0] @ offset 12
Produced record to topic test1 partition [0] @ offset 13
Produced record to topic test1 partition [0] @ offset 14
Produced record to topic test1 partition [0] @ offset 15
Produced record to topic test1 partition [0] @ offset 16
Produced record to topic test1 partition [0] @ offset 17
Produced record to topic test1 partition [0] @ offset 18
Produced record to topic test1 partition [0] @ offset 19
10 messages were produced to topic test1!
```

ybyzek:
@rspurgeon nice find.  What about the new warning introduced, can it be eliminated?

```
./producer.py:70: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
  p.produce(topic, key=record_key, value=record_value, on_delivery=acked)
```

rspurgeon:
> @rspurgeon nice find. What about the new warning introduced, can it be eliminated?
> 
> ```
> ./producer.py:70: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
>   p.produce(topic, key=record_key, value=record_value, on_delivery=acked)
> ```

Here is a related to confluent issue: https://github.com/confluentinc/confluent-kafka-python/issues/763

I'm not certain what version of python 3 gets installed in the ubuntu 18.04 example.  I can look into it further and I can try an older version of python 3

rspurgeon:
@ybyzek I do **not** get the deprecation warning on python 3.7

```
FROM python:3.7-slim

COPY requirements.txt /tmp/requirements.txt
RUN pip3 install -U -r /tmp/requirements.txt

COPY *.py ./

```

clients/cloud/python/README.md
```diff
@@ -197,3 +197,34 @@ Consumed record with key alice and value 9,                       and updated to
...
     ```
+
+# Example 3: Run All the Above in Docker
+
+You also may run all the above code from within Docker.
+
+1. Copy to the current directory your local file with configuration parameters to connect to your Confluent Cloud instance.
+
+```bash
+$ cp $HOME/.confluent/librdkafka.config .
```
rspurgeon: You could consider instructing the user to just mount the config file directly from their home directory below instead of adding this copy step. 

docker run -v $HOME/.confluent/librdkafka.config:/root/.confluent/librdkafka.config -it --rm cloud-demo-python

chuck-confluent:
I wonder if it might be a good idea to simplify the use of `librdkafka.config`. Maybe the source directory should include its own `librdkafka.config` file. Then `COPY . /app` and `WORKDIR /app`. This is nice because running a shell and doing `ls` will be tidier. Then you can drop the volume mount and do `-f librdkafka.config` instead of `-f $HOME/.confluent/librdkafka.config`. The downside is changes to the config file would mean having to rebuild the image (I suppose you could still overwrite it with a volume mount, but then you might forget to rebuild the image later and wonder why it isn't working properly later)

clients/cloud/python/README.md
ybyzek: @rspurgeon *much* better yes.

ybyzek:
Thanks @chuck-confluent 

> I wonder if it might be a good idea to simplify the use of `librdkafka.config`. Maybe the source directory should include its own `librdkafka.config` file. Then `COPY . /app` and `WORKDIR /app`. This is nice because running a shell and doing `ls` will be tidier. Then you can drop the volume mount and do `-f librdkafka.config` instead of `-f $HOME/.confluent/librdkafka.config`. The downside is changes to the config file would mean having to rebuild the image (I suppose you could still overwrite it with a volume mount, but then you might forget to rebuild the image later and wonder why it isn't working properly later)

I actually started this way with adding `librdkafka.config` into the Dockerfile and image build, but changed it for two reasons:

1. Docker context -- couldn't pull in file from parent directory
2. Conservative -- didn't want a Docker image with a user's credentials built into it, and then a user might accidentally share that image.

With the commit at https://github.com/confluentinc/examples/pull/602/commits/1a3f83de60d9501a78d475b5030dc7024feeafcc (with integrates feedback from @rspurgeon ), it eliminates a `cp` and is much smoother.  Added benefit: the commands are exactly the same in Docker and non-Docker

rspurgeon:
> Thanks @chuck-confluent
> 
> > I wonder if it might be a good idea to simplify the use of `librdkafka.config`. Maybe the source directory should include its own `librdkafka.config` file. Then `COPY . /app` and `WORKDIR /app`. This is nice because running a shell and doing `ls` will be tidier. Then you can drop the volume mount and do `-f librdkafka.config` instead of `-f $HOME/.confluent/librdkafka.config`. The downside is changes to the config file would mean having to rebuild the image (I suppose you could still overwrite it with a volume mount, but then you might forget to rebuild the image later and wonder why it isn't working properly later)
> 
> I actually started this way with adding `librdkafka.config` into the Dockerfile and image build, but changed it for two reasons:
> 
>     1. Docker context -- couldn't pull in file from parent directory
> 
>     2. Conservative -- didn't want a Docker image with a user's credentials built into it, and then a user might accidentally share that image.
> 
> 
> With the commit at [1a3f83d](https://github.com/confluentinc/examples/commit/1a3f83de60d9501a78d475b5030dc7024feeafcc) (with integrates feedback from @rspurgeon ), it eliminates a `cp` and is much smoother. Added benefit: the commands are exactly the same in Docker and non-Docker

+1 @ybyzek when I reviewed the previous version I was going to flag baking secrets into the image but then you revised it

.gitignore
```diff
@@ -20,3 +20,4 @@ log.*
 *.csr
 *.srl
 *_creds
+**/librdkafka.config
```
rspurgeon: don't necessarily need this anymore

rspurgeon (approved):
Minor comment remaining about .gitignore, otherwise LGTM.  I like this change a lot!

.gitignore
ybyzek: Ack and removed

chuck-confluent:
It still might be worth doing the `COPY *.py /app` and `WORKDIR /app` to make it a little cleaner when you bring up a shell in the container, but that's super nitpicky. This is great!

clients/cloud/python/requirements.txt
```diff
@@ -0,0 +1,7 @@
+requests
+fastavro>=0.23.0
+avro-python3==1.9.2.1
+jsonschema
+protobuf
+confluent-kafka
```
javabrett, line 6: This line is redundant, as base is included in the following line which includes extras `[avro]`.

javabrett, line 6: Since this isn't a version-locked environment file (nor should it be for this purpose), my temptation is to replace this entire file with the single entry:

```
confluent-kafka[avro,jsonschema,protobuf]
```
This installs the lib and the extras for schema variants, plus all the transitive dependencies using `confluent-kafka`'s `setup.py`.  It avoids mentioning transitive dependencies and versions which might in-future become non-dependencies or have different bounds.

If favoured, I can send a PR.

ybyzek:
@javabrett you have made some great suggestions that I wasn't aware could be done.  (Sorry to have merged so quickly.)  If you are amenable to making a PR, that would be fabulous, thank you.

PR #607, opened 2020-04-29
Improve CCloud CLI script

ybyzek:
Improvement by @rmoff

### DIFF
--- a/ccloud/beginner-cloud/start.sh
+++ b/ccloud/beginner-cloud/start.sh
@@ -109,9 +109,8 @@ if [[ $status != 0 ]]; then
 fi
 
 echo -e "\n# Produce 10 messages to topic $TOPIC1"
-echo '(for i in `seq 1 10`; do echo "${i}" ; done) | tee | \'
-echo "timeout 10s ccloud kafka topic produce $TOPIC1"
-(for i in `seq 1 10`; do echo "${i}" ; done) | tee | timeout 10s ccloud kafka topic produce $TOPIC1
+echo 'seq 1 10 | tee /dev/tty | ccloud kafka topic produce $TOPIC1'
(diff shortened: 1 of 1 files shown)


rspurgeon:
@rmoff could you provide some context for this change please?

ccloud/beginner-cloud/start.sh
```diff
@@ -109,16 +109,13 @@ if [[ $status != 0 ]]; then
 fi
 
 echo -e "\n# Produce 10 messages to topic $TOPIC1"
-echo '(for i in `seq 1 10`; do echo "${i}" ; done) | \'
-echo "timeout 10s ccloud kafka topic produce $TOPIC1"
-(for i in `seq 1 10`; do echo "${i}" ; done) | timeout 10s ccloud kafka topic produce $TOPIC1
```
rspurgeon, line 114: does the removal of the usage of `timeout` have a side effect that we need to consider?

rmoff: with the timeout in place the producer might exit before all messages are sent

rmoff:
> @rmoff could you provide some context for this change please?

I was going through the script for recording, and figured `tee` would fit better here to actually echo what's being sent to the producer, instead of faking it by sending data to the producer and then echoing to screen the output of the same statement that was used to generate the data. 

It's not a biggie and probably not worth the time being spent looking at it, so feel free to close :)

ccloud/beginner-cloud/start.sh
rspurgeon: Nice observation.  I'm wondering about the failure case, does the producer have it's own built in timeout to handle misconfigurations or timeouts for us so we don't just hang here indefinitely?

rspurgeon:
@rmoff closed this for now, we just wanted to test out the side effect of removing the timeout entirely.  Tracking in DEVX-1780

PR #609, opened 2020-04-30
DEVX-1770: CCloud & cloud-etl demo: pre-flight check for config file

ybyzek:


rspurgeon:
I ran this without an environment set and got the following, note that the demo continued after stop regardless of the stop step failing.  It proceeded to install datagen and run CP even though it was going to fail later.   Should we determine if stop.sh is failing properly and also move the preflight check up higher to fail faster?

```
[I] ➜ ./start.sh
====== Verifying prerequisites
Done

====== Cleaning up previous run
    The local commands are intended for a single-node development environment
    only, NOT for production usage. https://docs.confluent.io/current/cli/index.html

Using CONFLUENT_CURRENT: /var/folders/vc/_h2l6_yd0hbdk__7vz2rs4180000gp/T/confluent.SPkp3Xyc
Stopping control-center
control-center is [DOWN]
Stopping ksql-server
ksql-server is [DOWN]
Stopping connect
connect is [DOWN]
Stopping kafka-rest
kafka-rest is [DOWN]
Stopping schema-registry
schema-registry is [DOWN]
Stopping kafka
kafka is [DOWN]
Stopping zookeeper
zookeeper is [DOWN]
Deleting: /var/folders/vc/_h2l6_yd0hbdk__7vz2rs4180000gp/T/confluent.SPkp3Xyc
ERROR: could not determine if environment is set. Run 'ccloud environment list' andset 'ccloud environment use' and try again

Done with cleanup
====== Installing kafka-connect-datagen
Running in a "--no-prompt" mode

```

ybyzek:
@rspurgeon that failure shown in your output is from stop.sh, which is called from start.sh. 

stop.sh fails then continues.  If you continued your output (which is currently cut), it does fail and exit out of start.sh.

I've pushed a commit to fail faster in start.sh.  (It doesn't "fix" by exiting from stop.sh)

rspurgeon:
> @rspurgeon that failure shown in your output is from stop.sh, which is called from start.sh.
> 
> stop.sh fails then continues. If you continued your output (which is currently cut), it does fail and exit out of start.sh.
> 
> I've pushed a commit to fail faster in start.sh. (It doesn't "fix" by exiting from stop.sh)

Right, my question was if stop fails, should start continue?  Isn't a stop failure an indication of an issue?

ybyzek:
That is the right way to fix it, but outside scope of this PR.  It would require another Jira/PR to restructure all our demos for start.sh to fail if stop.sh fails.  For now what I implemented is "fail faster" in start.sh

rspurgeon (approved):
LGTM

PR #616, opened 2020-05-06
csharp .NET Confluent Cloud example client improvements

javabrett:
This PR originally motivated by the example code not being able to correctly parse `librdkafka.config` as suggested by docs.  `Enum.TryParse` is silently failing and returning false, as the config keys in `librdkafka.config` don't match the enum values.

`ClientConfig` has a dictionary constructor which is able to parse these values, so the fix is to just use that.

Along the way improvements (could be separately applied in a different PR if preferred, but I only tested with these in-place due to environment constraints):

- Added a `.gitignore` for `bin/`, `obj/` and `*.swp`.
- Updated to `netcoreapp3.1` and lang `8.0` so I could test this easily on Mac.
- Upgraded to latest GA `Confluent.Kafka` version `1.4.0`.
- Upgraded `Newtonsoft.Json` to current `12.0.3`.
- Fixed 2 LOC with new deprecation warnings with `Confluent.Kafka` upgrade.

Thanks @mhowlett for the dictionary tip.

### DIFF
--- a/clients/cloud/csharp/CCloud.csproj
+++ b/clients/cloud/csharp/CCloud.csproj
@@ -3,7 +3,6 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp3.1</TargetFramework>
-    <LangVersion>8.0</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
--- a/clients/cloud/csharp/CCloud.csproj
+++ b/clients/cloud/csharp/CCloud.csproj
@@ -2,8 +2,8 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
-    <LangVersion>7.3</LangVersion>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <LangVersion>8.0</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
--- a/clients/cloud/csharp/CCloud.csproj
+++ b/clients/cloud/csharp/CCloud.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Confluent.Kafka" Version="1.4.0" />
+    <PackageReference Include="Confluent.Kafka" Version="1.4.2" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
   </ItemGroup>
 
--- a/clients/cloud/csharp/CCloud.csproj
+++ b/clients/cloud/csharp/CCloud.csproj
@@ -7,8 +7,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Confluent.Kafka" Version="1.0.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="12.0.1" />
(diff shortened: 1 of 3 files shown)


clients/cloud/csharp/CCloud.csproj
```diff
@@ -2,13 +2,13 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
-    <LangVersion>7.3</LangVersion>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <LangVersion>8.0</LangVersion>
```
mhowlett: Pretty sure you can remove `LangVersion` if targeting 3.1.

mhowlett: (will default to 8.0)

clients/cloud/csharp/CCloud.csproj
```diff
@@ -2,13 +2,13 @@
...
-    <TargetFramework>netcoreapp2.1</TargetFramework>
-    <LangVersion>7.3</LangVersion>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <LangVersion>8.0</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Confluent.Kafka" Version="1.0.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="12.0.1" />
+    <PackageReference Include="Confluent.Kafka" Version="1.4.0" />
```
mhowlett: i'll release 1.4.2 today, which is worth waiting for in terms of stability. maybe merge this tomorrow with that change.

clients/cloud/csharp/Program.cs
```diff
@@ -37,16 +37,8 @@ static async Task<ClientConfig> LoadConfig(string configPath, string certDir)
...
-                var clientConfig = new ClientConfig
-                {
-                    BootstrapServers = cloudConfig["bootstrap.servers"].Replace("\\", ""),
-                    SaslMechanism = saslMechanism,
-                    SecurityProtocol = securityProtocol,
-                    SaslUsername = cloudConfig["sasl.username"],
-                    SaslPassword = cloudConfig["sasl.password"],
-                };
+
+                var clientConfig = new ClientConfig(cloudConfig);
```
mhowlett, line 41: ok, this is probably best. but some commentary:

fwiw, the issue is the sasl protocol name has underscores in it in the config file, e.g. `sasl_ssl`, but this doesn't follow C# naming conventions so our enum values don't follow the standard config value names (they remove the underscore), so parsing the enum string doesn't work. the required translation is implemented in the `Config` class, and if you use the `Set` method (or constructor as you're doing), it'll all just work. which is great, but it's not ideal for the example as it's complicated to explain, and we don't want to be highlighting features that people shouldn't be using, which we are here.

the root cause of my concerns with all this is the config file is in an inappropriate format for the C# examples.

anyway, we do demonstrate normal use with the `ProducerConfig` and `ConsumerConfig` later on, so i guess that's all good. still not happy about this though.

one more thing: using `Enum.TryParse` isn't appropriate here - it's intended to use in `if` statements, the try parse will just fail silently without looking at the return value.

mhowlett (approved):
thanks for fixing this! two minor things to change, but otherwise lgtm. pre-approval given.

PR #623, opened 2020-05-08
DEVX-1768: Add to cp-quickstart: create version with cp-all-in-one-cloud to Confluent Cloud

ybyzek:
To test:

```
git checkout DEVX-1768
cd cp-quickstart
./start-docker-cloud.sh
./stop-docker-cloud.sh stack-configs/java-service-account-<SERVICE ID>.config
```


caleb-grillo (approved):
Verified that this works on my machine. 

Suggested improvements for later iterations:
- Provide information about how much we expect this demo to cost per hour in the acknowledgement checkpoint. The inputs should be deterministic enough that we can do this relatively easily.
- Idea: Create a multi-use promo code that covers the cost of this demo running for 1 or 2 hours and provide it at the acknowledgement checkpoint at the beginning. "To run this quickstart for free for 2 hours, enter DEMO1234 in the Cloud UI." Could also put that in the readme or something to avoid overloading the checkpoint text.

ccloud/README.md
```diff
@@ -23,24 +29,16 @@ This repo has a few resources to help you validate your solutions on Confluent C
 All demos/scripts that connect to Confluent Cloud use real Confluent Cloud resources.
 To avoid unexpected charges, carefully evaluate the cost of resources before launching any demo and ensure all resources are destroyed after you are done running it.
 
-# Confluent Cloud CLI
+Do not run any demo against your production Confluent Cloud cluster.
+Some of these scripts create clusters, topics, resources, ACLs, service accounts, etc, and delete them as well, so run them only in a development cluster.
```
rspurgeon:
```suggestion
These scripts create environments, clusters, topics, ACLs, service accounts, applications, and other resources. The scripts provide functions to delete created resources; however, you should verify the deletion of all demo resources manually.

File: ccloud/README.md [hunk 2]
```diff
@@ -23,24 +29,16 @@ This repo has a few resources to help you validate your solutions on Confluent C
 All demos/scripts that connect to Confluent Cloud use real Confluent Cloud resources.
 To avoid unexpected charges, carefully evaluate the cost of resources before launching any demo and ensure all resources are destroyed after you are done running it.
 
-# Confluent Cloud CLI
+Do not run any demo against your production Confluent Cloud cluster.
```
rspurgeon:
```suggestion
These demos are meant for development environments only, do not run any demo against your production Confluent Cloud cluster.
```

cp-quickstart/stop-docker-cloud.sh
```diff
@@ -0,0 +1,15 @@
...
+if [ -z "$1" ]; then
+  echo "ERROR: Must supply argument that is the client configuration file created from './start.sh'. (Is it in stack-configs/ folder?) "
+  exit 1
+else
+  DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+  CONFIG_FILE=${DIR}/$1
+  ../ccloud/ccloud-stack/ccloud_stack_destroy.sh $CONFIG_FILE
+fi
+
+docker-compose down -v
```
rspurgeon: Maybe the docker-compose down should occur prior to the stack destroy.  Docker containers here are acting like clients, which probably should be destroyed before services.  They probably error when the service is taken out from under them, which might cause confusion.  

I'm not sure if docker-compose down should run even if the user erroneously calls the function without the config file parameter, but if they are calling stop, at least clients would stop pumping data to the services, reducing potential unexpected data flow up to cloud.  Calling docker-compose down twice once the user realizes they didn't provide the config the first time is idempotent.

ybyzek: @rspurgeon I went back and forth on this, but your last point is very compelling, and I switched this around (docker-compose down is first now).

rspurgeon (approved):
LGTM

PR #624, opened 2020-05-12
DEVX-1787: Adapt existing ccloud demo to use new ccloud-stack

ybyzek:


ccloud/start-docker.sh
```diff
@@ -4,42 +4,35 @@
...
-#   schema.registry.basic.auth.user.info=<SR API KEY>:<SR API SECRET>
-#   schema.registry.url=https://<SR ENDPOINT>
-#   # Confluent Cloud KSQL
-#   ksql.endpoint=https://<KSQL ENDPOINT>
-#   ksql.basic.auth.user.info=<KSQL API KEY>:<KSQL API SECRET>
-export CONFIG_FILE=~/.ccloud/config
-
+check_ccloud_version 1.0.0 \
+  && print_pass "ccloud version ok" \
+  || exit 1
```
rspurgeon, line 9: I don't believe these exit calls will ever be invoked given the design of the check_ccloud_version which will exit on version check failure prior to getting to this "or".  I think it's fine to make it like the following until the functions change to only return codes instead of exit the shell.

```suggestion
check_ccloud_version 1.0.0 \
  && print_pass "ccloud version ok"
```

rspurgeon (approved):
LGTM, tested local and docker versions with success, only minor feedback on some function usage.

ccloud/start-docker.sh
ybyzek: For consistency I will leave it as-is (because of how it's used in other scripts).

PR #649, opened 2020-06-02
adding ccloud::create_ccloud_stack_sub

gianlucanatali:
extracting logic from  ccloud::create_ccloud_stack into reusable function for my demo (ccloud::create_ccloud_stack_sub)

I have moved the definition of the variables up, so I can change them for my demo.
The name of the function is not great, happy to change that.
In general consider this a newbie contribution, so if anything sounds off let me know!!
Thank you @rspurgeon

### DIFF
--- a/utils/ccloud_library.sh
+++ b/utils/ccloud_library.sh
@@ -737,6 +737,16 @@ function ccloud::create_ccloud_stack() {
   CLUSTER_NAME=demo-kafka-cluster-$SERVICE_ACCOUNT_ID
   CLUSTER_CLOUD="${CLUSTER_CLOUD:-aws}"
   CLUSTER_REGION="${CLUSTER_REGION:-us-west-2}"
+  KSQL_NAME="demo-ksql-$SERVICE_ACCOUNT_ID"
+
+  mkdir -p stack-configs
+  CLIENT_CONFIG="stack-configs/java-service-account-$SERVICE_ACCOUNT_ID.config"
+
+  ccloud::create_ccloud_stack_sub
+}
+
+function ccloud::create_ccloud_stack_sub() {
+  
   CLUSTER=$(ccloud::create_and_use_cluster $CLUSTER_NAME $CLUSTER_CLOUD $CLUSTER_REGION)
   if [[ "$CLUSTER" == "" ]] ; then
     echo "Kafka cluster id is empty"
@@ -759,7 +769,6 @@ function ccloud::create_ccloud_stack() {
   SCHEMA_REGISTRY_CREDS=$(ccloud::create_credentials_resource $SERVICE_ACCOUNT_ID $SCHEMA_REGISTRY)
 
   if $enable_ksql ; then
-    KSQL_NAME="demo-ksql-$SERVICE_ACCOUNT_ID"
     KSQL=$(ccloud::create_ksql_app $KSQL_NAME $CLUSTER)
     KSQL_ENDPOINT=$(ccloud ksql app describe $KSQL -o json | jq -r ".endpoint")
     KSQL_CREDS=$(ccloud::create_credentials_resource $SERVICE_ACCOUNT_ID $KSQL)
@@ -768,8 +777,7 @@ function ccloud::create_ccloud_stack() {
 
   ccloud::create_acls_all_resources_full_access $SERVICE_ACCOUNT_ID
 
-  mkdir -p stack-configs
-  CLIENT_CONFIG="stack-configs/java-service-account-$SERVICE_ACCOUNT_ID.config"
+  
   cat <<EOF > $CLIENT_CONFIG
 # ------------------------------
 # Confluent Cloud connection information for demo purposes only
--- a/utils/ccloud_library.sh
+++ b/utils/ccloud_library.sh
@@ -721,6 +721,16 @@ function ccloud::create_ccloud_stack() {
   CLUSTER_NAME=demo-kafka-cluster-$SERVICE_ACCOUNT_ID
   CLUSTER_CLOUD="${CLUSTER_CLOUD:-aws}"
   CLUSTER_REGION="${CLUSTER_REGION:-us-west-2}"
+  KSQL_NAME="demo-ksql-$SERVICE_ACCOUNT_ID"
+
+  mkdir -p stack-configs
+  CLIENT_CONFIG="stack-configs/java-service-account-$SERVICE_ACCOUNT_ID.config"
+
+  ccloud::create_ccloud_stack_sub
+}
+
+function ccloud::create_ccloud_stack_sub() {
+  
   CLUSTER=$(ccloud::create_and_use_cluster $CLUSTER_NAME $CLUSTER_CLOUD $CLUSTER_REGION)
   if [[ "$CLUSTER" == "" ]] ; then
     print_error "Kafka cluster id is empty"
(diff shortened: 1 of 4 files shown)


rspurgeon:
@gianlucanatali I believe I understand the goal.  I think what I would prefer we do is make the variables in the existing function overridable instead of creating a new function entirely.

For example, if your goal is to be able to set KSQL_NAME from a call to `ccloud::create_ccloud_stack`, then we could assign `KSQL_NAME` with an optional default like so:

`KSQL_NAME=${KSQL_NAME:-"demo-ksql-$SERVICE_ACCOUNT_ID"}`

The same is true with `CLIENT_CONFIG`, if you desire to set that from your calling function, we can just setup a default like above.

I'm curious why you want to remove the `mkdir -p stack-configs` from the function that you're calling, are you storing the configs elsewhere?  If so, we could do a similar thing and make the path to the configs configurable with a default.

Thanks for your PR and let me know what you think about these proposed changes.  If you'd like some help w/ the bash code, please let me know.

gianlucanatali:
Hi @rspurgeon , thanks for quick feedback. Actually I am only using part of this function (the part I isolated in that function). I enter this function having already the environment and service account created. In my case I have 1 env, 1 service account, and 3 stacks inside. You can get a idea by looking at this code where I use the function:
https://github.com/gianlucanatali/confluent-multicloud-demo/blob/b8b8bc597adfbe35361f531c666cfb62f79013ea/create_demo_environment.sh#L218

This is thw starting point, here I actually use the function:
https://github.com/gianlucanatali/confluent-multicloud-demo/blob/b8b8bc597adfbe35361f531c666cfb62f79013ea/create_demo_environment.sh#L207

Thanks!

utils/ccloud_library.sh
```diff
@@ -721,6 +721,16 @@ function ccloud::create_ccloud_stack() {
...
   CLUSTER_REGION="${CLUSTER_REGION:-us-west-2}"
+  KSQL_NAME="demo-ksql-$SERVICE_ACCOUNT_ID"
+
+  mkdir -p stack-configs
+  CLIENT_CONFIG="stack-configs/java-service-account-$SERVICE_ACCOUNT_ID.config"
+
+  ccloud::create_ccloud_stack_sub
+}
+
+function ccloud::create_ccloud_stack_sub() {
```
ybyzek: Option 1: use the sub function but rename it to something like `create_ccloud_substack_per_env` ? (This may be the simplest option...just requires picking a good name)

Option 2: remove the entire idea of a sub function, and instead refactor `create_ccloud_stack` to use options, thus allowing user to pass in pre-existing service account id and environment to use.  (This is more complicated upfront but might give more freedom in the future)

cc: @rspurgeon WDYT?

gianlucanatali:
Thanks @ybyzek and @rspurgeon fo help and guidance, I have changed the function using Option 2 as also discussed with Rick via slack, this seems a better approach.
I did not test the existing demo assets (other than mine) with this new version though.
Let me know what you think!

utils/ccloud_library.sh
rspurgeon: As @gianlucanatali mentioned below, we discussed out of band making the function default variables instead of breaking the function up.  He's going to work on a new verison and push a change for that.

utils/ccloud_library.sh
```diff
@@ -708,29 +708,26 @@ function ccloud::set_kafka_cluster_use() {
...
+    # Service Account is not received so it will be created
+    local RANDOM_NUM=$((1 + RANDOM % 1000000))
+    SERVICE_NAME=${SERVICE_NAME:-"demo-app-$RANDOM_NUM"}
+    SERVICE_ACCOUNT_ID=$(ccloud::create_service_account $SERVICE_NAME)
+    
+  fi
 
-  SERVICE_NAME="demo-app-$RANDOM_NUM"
-  SERVICE_ACCOUNT_ID=$(ccloud::create_service_account $SERVICE_NAME)
   echo "Creating Confluent Cloud stack for new service account id $SERVICE_ACCOUNT_ID of name $SERVICE_NAME"
```
rspurgeon:
```suggestion
  echo "Creating Confluent Cloud stack for service account $SERVICE_NAME, ID: $SERVICE_ACCOUNT_ID.
```

utils/ccloud_library.sh
```diff
@@ -708,29 +708,26 @@ function ccloud::set_kafka_cluster_use() {
...
 
-  ENVIRONMENT_NAME="demo-env-$SERVICE_ACCOUNT_ID"
-  ENVIRONMENT=$(ccloud::create_and_use_environment $ENVIRONMENT_NAME)
-
-  CLUSTER_NAME=demo-kafka-cluster-$SERVICE_ACCOUNT_ID
+  if [[ -z "$ENVIRONMENT" ]]; then
+    # Environment is not received so it will be created
+    echo "You need to specify a demo Type!"
```
rspurgeon: I'm not sure what this print line is specifying?  It doesn't seem necessary

utils/ccloud_library.sh
```diff
@@ -708,29 +708,26 @@ function ccloud::set_kafka_cluster_use() {
...
   enable_ksql=$1
 
-  RANDOM_NUM=$((1 + RANDOM % 1000000))
-  #echo "RANDOM_NUM: $RANDOM_NUM"
+  if [[ -z "$SERVICE_ACCOUNT_ID" ]]; then
+    # Service Account is not received so it will be created
+    local RANDOM_NUM=$((1 + RANDOM % 1000000))
+    SERVICE_NAME=${SERVICE_NAME:-"demo-app-$RANDOM_NUM"}
+    SERVICE_ACCOUNT_ID=$(ccloud::create_service_account $SERVICE_NAME)
+
```
rspurgeon: nit: remove empty line

utils/ccloud_library.sh
```diff
@@ -708,29 +708,26 @@ function ccloud::set_kafka_cluster_use() {
...
 
-  RANDOM_NUM=$((1 + RANDOM % 1000000))
-  #echo "RANDOM_NUM: $RANDOM_NUM"
+  if [[ -z "$SERVICE_ACCOUNT_ID" ]]; then
+    # Service Account is not received so it will be created
+    local RANDOM_NUM=$((1 + RANDOM % 1000000))
+    SERVICE_NAME=${SERVICE_NAME:-"demo-app-$RANDOM_NUM"}
+    SERVICE_ACCOUNT_ID=$(ccloud::create_service_account $SERVICE_NAME)
+    
+  fi
```
rspurgeon, line 732: After this line, I believe that we should check for empty SERVICE_NAME and exit with a message if it's not provided or wasn't calculated in the case of the SERVICE_ACCOUNT_ID being provided.

rspurgeon:
@gianlucanatali Please rebase off 5.5.0-post to pickup some important changes related to these functions

gianlucanatali:
Hi @rspurgeon thanks a lot for your comments, I'll go ahead and change the function accordingly. I am not sure what I did wrong... I rebased off 5.5.0-post , and I now see on the changed files of this PR also files I don't think I changed ... Can you have a look?
I can also create a new PR and just bring with me the code in another branch but I think I need to figure out how to do this properly at some point :) Thanks!

utils/ccloud_library.sh
gianlucanatali: yes, good catch, was a copy paste refuse. I will remove it

utils/ccloud_library.sh
gianlucanatali: ✅

utils/ccloud_library.sh
gianlucanatali: Added the logic in, thx!

utils/ccloud_library.sh
gianlucanatali: ✅

gianlucanatali:
I pushed the changes to my branch, as I mentioned my intent is only to change the function function ccloud::create_ccloud_stack()  in ccloud_library.sh . But I do see other files changed in this PR, probably I screwed smg up in the rebase process.. guidance welcome :)
Thanks!

gianlucanatali:
Closing this PR , moving same changes to new one: https://github.com/confluentinc/examples/pull/656

PR #650, opened 2020-06-02
DEVX-1838: Ensure pageviews topic and subject exist before running KSQL

ybyzek:


ybyzek:
@markteehan validated this PR fixes the demo run in his environment.  Merging now

PR #651, opened 2020-06-03
Removes a missing function from ccloud create stack

rspurgeon:
When the utils/helper.sh was removed as a source for this file
this function no longer became available.


ybyzek (commented):
@rspurgeon nice catch, LGTM

PR #652, opened 2020-06-03
Quick fix for quoted params

rspurgeon:
Kafka Cluster and ksqlDB app name can contain spaces, so quoting the variables at the execution point of the command allows this.

Before this, names with spaces would have produced an error.


ybyzek (approved):
LGTM

PR #603, opened 2020-04-28
Simplified requirements.txt file to mention only confluent-kafka+extras

javabrett:
Dependencies will be automatically installed by pip.

PR #657, opened 2020-06-04
DEVX-1826: Updates python example to Serialization API

rspurgeon:


clients/cloud/python/consumer.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
rspurgeon: I'm not sure about this change.  On my system, where I have a python 2.7 installation, `env python` resolves to python 2, and cannot locate the confluent-kafka package or run the examples.  I'm not sure of the recommended method for making this work on a system like mine, or if it should be a system specific configuration change instead of this.  @rnpridgeon can you help me if this is a valid change or should users be doing something local that I can document?

clients/cloud/python/README.md
```diff
@@ -5,7 +5,7 @@ Produce messages to and consume messages from a Kafka cluster using [Confluent P
 
 # Prerequisites
 
-* [Confluent's Python Client for Apache Kafka](https://github.com/confluentinc/confluent-kafka-python) installed on your machine. Check that you are using version 1.0.0 or higher (e.g., `pip show confluent-kafka`).
+* [Confluent's Python Client for Apache Kafka](https://github.com/confluentinc/confluent-kafka-python) installed on your machine. Check that you are using version 1.4.2 or higher (e.g., `pip show confluent-kafka`).
```
ybyzek: Python2 is no longer supported, so propose

```suggestion
* [Confluent's Python Client for Apache Kafka](https://github.com/confluentinc/confluent-kafka-python) installed on your machine. Check that you are using version 1.4.2 or higher (e.g., `pip3 show confluent-kafka`).
```

clients/cloud/python/consumer.py
ybyzek: I defer to @rnpridgeon but I believe users  often alias `python` to their local `python2` or `python3`, so consider leaving as `python`.

ybyzek (approved):
LGTM pending approval from @confluentinc/clients

ybyzek:
@rspurgeon did you validate the Docker-based Python example and determine whether https://github.com/confluentinc/examples/blob/5.5.0-post/clients/cloud/python/requirements.txt also needs an update?

clients/cloud/python/consumer.py
rspurgeon: IIUC `env` does not reference aliases

clients/cloud/python/README.md
rspurgeon: This is similar to the `python` shebang issue above.  I'm not familiar enough with the python ecosystem.  If we change the env above, I think this makes sense stating that we only support python3, but if we don't change above, then this seems confusing.

clients/cloud/python/consumer.py
rspurgeon: In my case I used pip3 (aliased on my machine from pip to pip3) to install confluent-kafka, but python 2 is found first in my path. I think this causes the issue I see when trying to run the scripts directly.  I assume others might experience this, so outside of uninstalling python2, adjusting my PATH order, or also using pip 2 to install confluent-kafka, is there a better remedy?

rspurgeon:
> @rspurgeon did you validate the Docker-based Python example and determine whether https://github.com/confluentinc/examples/blob/5.5.0-post/clients/cloud/python/requirements.txt also needs an update?

@ybyzek the Docker instructions work.   When building the docker image the following warnings are given, but the applications work as expected:

```
WARNING: confluent-kafka 1.4.2 does not provide the extra 'jsonschema'
...
WARNING: You are using pip version 20.0.2; however, version 20.1.1 is available.
You should consider upgrading via the '/usr/local/bin/python -m pip install --upgrade pip' command
```

The addition of `jsonschema` was made by @javabrett in https://github.com/confluentinc/examples/pull/603

clients/cloud/python/README.md
rnpridgeon: Virtualenv should be used rather than relying on the system installation. This will simplify things as you can explicitly instruct it to run a particular interpreter. 

https://virtualenv.pypa.io/en/latest/user_guide.html#introduction

clients/cloud/python/consumer.py
rnpridgeon: see comment about virtualenv

clients/cloud/python/README.md
rspurgeon: 🤦 , of course... Thanks @rnpridgeon .  I'm going to add some documentation to help users if this is new to them or have memory failure like I do.

rspurgeon: I've added some instructions for installing and running Virtualenv to the README

rspurgeon:
@confluentinc/clients Could someone provide a review for this?  Thank you

clients/cloud/python/README.md
```diff
@@ -103,13 +106,14 @@ Waiting for message or event/error in poll()
...
 ```
 
-
 # Example 2: Avro And Confluent Cloud Schema Registry
 
 This example is similar to the previous example, except the key and value are formatted as Avro and integrates with the Confluent Cloud Schema Registry.
 Before using Confluent Cloud Schema Registry, check its [availability and limits](https://docs.confluent.io/current/cloud/limits.html?utm_source=github&utm_medium=demo&utm_campaign=ch.examples_type.community_content.clients-ccloud).
 Note that your VPC must be able to connect to the Confluent Cloud Schema Registry public internet endpoint.
 
+These examples utilize the latest Serializer API provided by the `confluent-kafka` library.  The Serializer API replaces the legacy AvroProducer and AvroConsumer classes to provide a more flexible API including additional support for JSON, Protobuf, and Avro data formats.  See the latest [confluent-kafka documentation](https://docs.confluent.io/current/clients/confluent-kafka-python/#) for further details.
```
mhowlett: the new apis are 'experimental and subject to change', is the main thing. In practice, they're essentially stable, except that I hope they'll be combined with the Producer and Consumer classes (not exposed as separate classes), and the details of that aren't completely clear.

i'm not sure how best to communicate this.

mhowlett: Probably don't want/need the `#` in the url.

clients/cloud/python/ccloud_lib.py
```diff
@@ -53,18 +52,24 @@ def __init__(self, name=None):
...
+
     def to_dict(self):
         """
             The Avro Python library does not support code generation.
             For this reason we must provide a dict representation of our class for serialization.
         """
-        return {
-            "name": self.name
-        }
+        return dict(name=self.name)
```
mhowlett, line 68: is there a reason to change the literal to a dict constructor? I'd use a literal and put it on one line `return { "name": self.name }`. but kinda bike shedding.

clients/cloud/python/ccloud_lib.py
```diff
@@ -89,14 +95,20 @@ def __init__(self, count=None):
...
+
     def to_dict(self):
         """
             The Avro Python library does not support code generation.
             For this reason we must provide a dict representation of our class for serialization.
         """
-        return {
-            "count": self.count
-        }
+        return dict(count=self.count)
```
mhowlett, line 111: same same

clients/cloud/python/ccloud_lib.py
rspurgeon: I modeled these changes after the official examples: https://github.com/confluentinc/confluent-kafka-python/blob/master/examples/avro_consumer.py#L62

mhowlett (approved):
LGTM, except i'm unsure how or if to communicate the API is marked as experimental.

rspurgeon:
> LGTM, except i'm unsure how or if to communicate the API is marked as experimental.

Thanks @mhowlett , I'm going to update the URL and push that minor change.   Regading the API stability, maybe if the official docs are updated with verbiage, we can copy and/or point to that. https://docs.confluent.io/current/clients/confluent-kafka-python/ (no mention of experimental Serialization API, at least on this front page).

clients/cloud/python/ccloud_lib.py
```diff
@@ -127,7 +139,7 @@ def read_ccloud_config(config_file):
     with open(config_file) as fh:
         for line in fh:
             line = line.strip()
-            if line[0] != "#" and len(line) != 0:
+            if len(line) != 0 and line[0] != "#":
```
ybyzek, line 142: Thank you!  ref: https://github.com/confluentinc/examples/issues/661#issuecomment-640745841

PR #659, opened 2020-06-05
Adds some missing functions to ccloud_library

rspurgeon:
For now these are duplicated out of helper.sh to maintain the
scripts independence.  We can resolve this duplication over time
if necessary, but this is an attempt to get the script usable
on its own

### DIFF
--- a/utils/ccloud_library.sh
+++ b/utils/ccloud_library.sh
@@ -46,7 +46,14 @@ function ccloud::prompt_continue_ccloud_demo() {
 
   return 0
 }
+function ccloud::validate_expect_installed() {
+  if [[ $(type expect 2>&1) =~ "not found" ]]; then
+    echo "'expect' is not found. Install 'expect' and try again"
+    exit 1
+  fi
 
+  return 0
+}
 function ccloud::validate_ccloud_cli_installed() {
   if [[ $(type ccloud 2>&1) =~ "not found" ]]; then
     echo "'ccloud' is not found. Install Confluent Cloud CLI (https://docs.confluent.io/current/quickstart/cloud-quickstart/index.html#step-2-install-the-ccloud-cli) and try again"
@@ -566,7 +573,7 @@ function ccloud::login_ccloud_cli(){
   EMAIL=$2
   PASSWORD=$3
 
-  ccloud::check_expect
+  ccloud::validate_expect_installed
 
   echo -e "\n# Login"
   OUTPUT=$(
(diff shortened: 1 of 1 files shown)


utils/ccloud_library.sh
```diff
@@ -88,7 +88,7 @@ function ccloud::validate_version_ccloud_cli() {
   REQUIRED_CCLOUD_VER=${1:-"0.185.0"}
   CCLOUD_VER=$(ccloud::get_version_ccloud_cli)
 
-  if version_gt $REQUIRED_CCLOUD_VER $CCLOUD_VER; then
+  if ccloud::version_gt $REQUIRED_CCLOUD_VER $CCLOUD_VER; then
```
ybyzek, line 98: While we're here, can we update 2 lines above

```
  REQUIRED_CCLOUD_VER=${1:-"0.185.0"}
```

to 1.7.0?

utils/ccloud_library.sh
```diff
@@ -836,3 +836,42 @@ function ccloud::destroy_ccloud_stack() {
...
+            curr_wait=$((curr_wait+sleep_interval))
+            sleep $sleep_interval
+        fi
+    done
+    printf "\n"
+}
+function ccloud::version_gt() { 
+  test "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1";
+}
+function ccloud::check_expect() {
```
ybyzek: In this library, all functions that started with `check` were intentionally converted to `validate`.  If this function is being brought into this library, it also has to change to follow the other functions that validate a bin is installed.  So:

`check_expect` --> `validate_expect_installed`

utils/ccloud_library.sh
```diff
@@ -517,7 +517,7 @@ function ccloud::wait_for_connector_up() {
 
   connectorName=$(cat $filename | jq -r .name)
   echo "Waiting up to $maxWait seconds for connector $filename ($connectorName) to be RUNNING"
-  retry $maxWait ccloud::validate_connector_up $connectorName || exit 1
+  ccloud::retry $maxWait ccloud::validate_connector_up $connectorName || exit 1
```
ybyzek, line 527: In order differentiate from the function in the other library, WDYT about renaming this

from `ccloud::retry` to `ccloud::retry_until`?

rspurgeon: It is distinquished by the pseudo namespace.  If we rename the function in one place but not in the other does it seem to a reader it does something different?

ybyzek: Good point.  It's just a weird situation having the same functionality in two different places, but current implementation is probably the best path forward, thanks!

ybyzek (approved):
LGTM thanks @rspurgeon

PR #660, opened 2020-06-08
DEVX-1825 k8s updates operator docker image

rspurgeon:
for the client-console pod. For now this is statically defined
until the best templating or environment solution is determined

### DIFF
--- a/kubernetes/gke-base/cfg/client-console-pod.yaml
+++ b/kubernetes/gke-base/cfg/client-console-pod.yaml
@@ -6,7 +6,7 @@ metadata:
 spec:
   containers:
   - name: client-console
-    image: docker.io/confluentinc/cp-server-operator:5.4.1.0
+    image: docker.io/confluentinc/cp-server-operator:5.5.0.0
     command: [sleep, "86400"]
     volumeMounts:
     - name: kafka-client-properties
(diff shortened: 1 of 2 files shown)


kubernetes/gke-base/cfg/client-console-pod.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
 spec:
   containers:
   - name: client-console
-    image: docker.io/confluentinc/cp-server-operator:5.4.1.0
+    image: docker.io/confluentinc/cp-server-operator:5.5.0.0
```
ybyzek, line 9: Should this be parameterized as `OPERATOR_CP_IMAGE_TAG`?

rspurgeon: @ybyzek please see the comment in the PR description and git commit msg.

IIUC it is not straightfoward to use a env var in a K8s definition or on the CLI with `kubectl`.  I'm investigating, but wanted to get the branch caught up for now.

ybyzek: Ack. Unless it can be tied to the parameter, does that mean we will always be playing catchup?

rspurgeon: It just means we have to manually manage this until I determine a way to leverage the config.env value.  I'm investigating and will update to that in another PR, and I won't close this issue.  I just wanted to get the branch "up to date" with this.

ybyzek (approved):
LGTM

PR #670, opened 2020-06-11
Update the clickstream demo to work with ksqlDB 0.10.0

big-andy-coates:
(0.10.0 which will be part of CP 6.0.0).

Tasks done:
* updated the SQL in `statements.sql` to be compliant with latest syntax
* restructured `statements.sql` to make it easier to follow (hopefully)
* Changed the `CLICK_USER_SESSIONS` table to track `WINDOWEND` as the timestamp that ES will use, as this means each new event has a new timestamp, allowing:
* Changed the Grafana dashboard so that the User Sessions table actually shows the events-per-session over time. (See screen shots below).
* Updated instructions slightly, including using ksqlDB `PRINT` to inspect topics, and adding an image of what it should look like.

### New dashboard shot included in the instructions (which show new session data too):

I always found the 'User sessions' graph very confusing. It was supposed to be counting the number of events within the session over time, yet it always looked mostly flat, with the occasional big spike.  On investigating, this turned out to be for two reasons:
 1. The underlying table was using `WINDOWSTART` time as the time of the event within ES, which means as a session grew, and the event count increased, ES was either overwriting the previous count, (as the timestamp matched), or ignoring them (as they were to far in the past).  Switching to using `WINDOWEND` fixed this.
 2. The Grafana dashboard graph was graphing the 'count' of rows in the top, not looking at the `EVENTS` count that ksqlDB was calculating.  So fix this.

The graph now tracks the number of events within the session over time:

New dashboard shot

### Old 'Sessions' dashboard shot:

Always found the `User Sessions` graph very confusing:

Old sessions shot

### New 'Sessions' dashboard shot:

New version, IMHO, clearly shows the sessions running over a period of time:

New sessions shot


sdandu-gh (approved):
LGTM, i walked through the demo step-by-step with the updated instructions with recent 6.0.x images and compared new grafana charts as well.

PR #672, opened 2020-06-11
Update cp-quickstart to latest ksqlDB syntax

big-andy-coates:
**NOT TESTED**

However, updated with the latest syntax.


sdandu-gh (approved):
LGTM, ran docker based cp-quickstart and confirmed tables and streams looked correct (ran a few queries, etc..)

PR #673, opened 2020-06-11
Update Microservice-orders to latest ksqlDB syntax

big-andy-coates:
**NOT TESTED**: as docker-compose up not working on 6.0.x or 5.5.x branches, and non-docker version requires to many hoops to be jumped through.

However, have updated the SQL statements inline with the latest syntax. Should be good to go.

Note: the `orders_cust1_joined` stream will now have the `customerid` in the key, not the value. Likewise, the `FRAUD_ORDER` table will have the `CUSTOMERID`, `LASTNAME`, `FIRSTNAME` in the key, not the value.   You can use `AS_VALUE` to copy into the value if needed.  See example in https://github.com/confluentinc/examples/pull/671 for more info.

### DIFF
--- a/microservices-orders/docker-compose.yml
+++ b/microservices-orders/docker-compose.yml
@@ -270,7 +270,6 @@ services:
       - "5601:5601"
     environment:
       xpack.security.enabled: "false"
-      XPACK_SECURITY_ENABLED: "false"
       xpack.monitoring.enabled: "false"
       discovery.type: "single-node"
       elasticsearch.url: http://elasticsearch:9200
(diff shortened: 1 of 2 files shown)


sdandu-gh (approved):
Ran both the archive and docker version on 6.0.x with the tutorial documentation and everything worked fine (one minor thing i'll create a separate PR for - unrelated to KSQL)

ybyzek:
@big-andy-coates thanks for the original PR!

@sdanduConf thanks for validating!

PR #675, opened 2020-06-11
Tweak promo code

ybyzek:


vsyu (commented):
If we want to make it more subtle, we can instead an asterisk after the period that links to https://www.confluent.io/confluent-cloud-promo-disclaimer/. It'd be nice to add a UTM code too. I've been using this format: ch.oprah_type.community_content.cloud.

ybyzek:
@vsyu thanks for the suggestions.

1. Agree with subtlety.  An asterisk may be missed but I will soften `disclaimer` to `details`
2. This is already in docs, so no UTM is required.

alelum (approved):
LGTM! Thank you!

PR #676, opened 2020-06-12
Update to Music example to latest ksqlDB syntax

big-andy-coates:
**SORT OF TESTED**: ran into issues testing using docker, so just pulled schemas across to local ksqlDB cluster and ran through examples. So while this PR includes the changes to the SQL needed to work with the latest syntax, I'm not convinced there aren't issues elsewhere.  The Music example could do with some more testing!


sdandu-gh (approved):
LGTM. I ran this demo on recent 6.0.x images and it worked.  A couple notes: 1) i ran with 5.5 kafka-streams-examples due to a 6.0.x bug, will run end-to-end with new images since it has been fixed, but shouldn't block this KSQL change 2) One note, @big-andy-coates , when creating ksql_songplays which joins with ksql_song table, i had too set to auto.offset earliest in the UI (I assume since the table has no new values), to make sure songplays didn't have null for album, artist, etc.. - i assume that is expected?

big-andy-coates:
@sdanduConf not sure that sounds 100% right.

By default, any events in the stream will be ignored when you create the join.  However, I would have thought this would result no output, not output with one side missing.

You can either produce play events after creating the join, or set `auto.offset.reset` to `earliest`.

Would be great if you could merge this and then make any follow up changes needed to address this :D

PR #677, opened 2020-06-12
DEVX-1865 silences some feedback output from ccloud

rspurgeon:
The objective here is to get the demo output back to where it was before
the change to nudge users for feedback from ccloud CLI.  We can expand
on the QUIET and redirection method over time if useful.

### DIFF
--- a/cp-quickstart/start-docker-cloud.sh
+++ b/cp-quickstart/start-docker-cloud.sh
@@ -6,6 +6,10 @@
 #########################################
 
 NAME=`basename "$0"`
+QUIET="${QUIET:-true}"
+[ -z $QUIET ] && 
+  REDIRECT_TO="/dev/stdout" ||
+  REDIRECT_TO="/dev/null"
 
 # Source library
 source ../utils/helper.sh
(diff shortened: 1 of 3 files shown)


cp-quickstart/start-docker-cloud.sh
```diff
@@ -6,6 +6,10 @@
 #########################################
 
 NAME=`basename "$0"`
+QUIET="${QUIET:-true}"
+[ -z $QUIET ] &&
```
ybyzek: @rspurgeon not sure if I'm interpreting this correctly, but given the line

```
QUIET="${QUIET:-true}"
```

Is it the case that 

1. `QUIET` will never have a non-zero length? (i.e., `-z` will always fail)
2. Due to above, `REDIRECT_TO` will always go to `/dev/null`?

Can you please elaborate on the desired behavior?

rspurgeon: Yes, that was confusing.  I think my original intent was for the user to turn off "QUIET" with QUIET="" but that doesn't seem correct.  I changed it in latest push to be if QUIET=="true" otherwise...

ybyzek (approved):
LGTM

PR #681, opened 2020-06-17
DEVX-1871: microservices: tidy and refactor to prepare for cloud

rspurgeon:
The general purpose of this change is to reduce duplication, standardize some usage
in docker-compose, and make some functions avaialable for use targeting other platforms
(ccloud).

Additionally, this change moves some sqlite code out of the shared utils folder and into
the microservices folder.  It changes the behavior of writing the database file to a
system level folder preferring modified files in local folders.

Connector configs now share a single template and are rendered to a local folder before
being posted to a connector cluster. This should reduce duplciation as well as allow for
posting from ccloud shortly.

Net result is 50 deletions with this change!  There are still more duplications that can be removed w/ the centralization of some behaviors into scripts, however the change was growing rather large and both demos are in a stable state at this point.


microservices-orders/start.sh
```diff
@@ -10,17 +10,26 @@ check_netstat || exit 1
...
 confluent-hub install --no-prompt confluentinc/kafka-connect-elasticsearch:$CONFLUENT
-echo "auto.offset.reset=earliest" >> $CONFLUENT_HOME/etc/ksqldb/ksql-server.properties
+grep -qxF 'auto.offset.reset=earliest' $CONFLUENT_HOME/etc/ksqldb/ksql-server.properties || echo 'auto.offset.reset=earliest' >> $CONFLUENT_HOME/etc/ksqldb/ksql-server.properties 
 confluent local start
 sleep 5
 
+export BOOTSTRAP_SERVER=localhost:9092
+export SCHEMA_REGISTRY_URL=http://localhost:8081
+export SQLITE_DB_PATH=${PWD}/db/data/microservices.db
+export ELASTICSEARCH_URL=http://localhost:9300
```
ybyzek: 9300: should this be 9200?

microservices-orders/scripts/render-connector-config.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
ybyzek, line 1: Why write the file to the system instead of just passing it in?

microservices-orders/start.sh
```diff
@@ -10,17 +10,26 @@ check_netstat || exit 1
...
+check_sqlite3 || exit 1
 
 ./stop.sh
 
 get_and_compile_kafka_streams_examples || exit 1
 
 confluent-hub install --no-prompt confluentinc/kafka-connect-jdbc:$CONFLUENT
 confluent-hub install --no-prompt confluentinc/kafka-connect-elasticsearch:$CONFLUENT
-echo "auto.offset.reset=earliest" >> $CONFLUENT_HOME/etc/ksqldb/ksql-server.properties
+grep -qxF 'auto.offset.reset=earliest' $CONFLUENT_HOME/etc/ksqldb/ksql-server.properties || echo 'auto.offset.reset=earliest' >> $CONFLUENT_HOME/etc/ksqldb/ksql-server.properties
```
ybyzek, line 21: This leaves one place for error: where the properties file has `auto.offset.reset=earliest` but is then overwritten later in the properties file with another value.

ybyzek (commented):
Nice improvements @rspurgeon .  I haven't validated this yet (presumably you have), but left a few comments in the meanwhile.

microservices-orders/start.sh
rspurgeon: I agree and thought of the situation where a user has `auto.offset.reset=latest` and then this will append the 'auto.offset.reset=earliest' line.  The previous code resulted in my properites file had many rows of the same entry after running the demo.  Any thoughts on handling these various cases?

microservices-orders/scripts/render-connector-config.sh
rspurgeon: There are three tools (curl, ccloud, and confluent local) that need to 'slurp' these values.  All of which do support direct slurping from files.  I found writing and reading files was the most straightforward way to template and pass the values to the various tools without having to deal with things like escaping values, dealing with shell issues, etc..  If you have have proposed changes that are more straightforward I'm very open to a suggested change here.

ybyzek: Would this work? https://github.com/confluentinc/examples/blob/5.5.0-post/utils/ccloud_library.sh#L506-L509

rspurgeon: After more thought, I still prefer the files.  As an artifcat of running the demo it's something that we could point to for educational material and be referenced later by the demo runner.  Keeping this script independent of the tool that slurps the values seems helpful IMO.  It also may prove useful to have the file allowing a user to more easily run the command themselves directly (ccloud connector config -- -d file.config) after they run the demo.

microservices-orders/start.sh
rspurgeon: yes, thank you... resolved

microservices-orders/scripts/render-connector-config.sh
ybyzek: Ok let's try it out ;)

rspurgeon:
@ybyzek I updated this to include using a script for starting and stopping the microservices java apps for both demos.  Please review when you have time, thank you

ybyzek (changes requested):
Just to note here, need to resolve render of empty param

```
    "connection.url": "",
```

rspurgeon:
> Just to note here, need to resolve render of empty param
> 
> ```
>     "connection.url": "",
> ```

Thank you very much for catching this @ybyzek .  I've addressed this and another item w/ ksqlDB when ran locally.  Thanks for your review.

microservices-orders/docker-compose.yml
```diff
@@ -185,69 +176,76 @@ services:
...
+        cub kafka-ready -b broker:9092 1 40
+        echo "Kafka ready"
+        echo "Creating demo topics"
+        /opt/docker/scripts/create-topics.sh
+        echo "Waiting 90 seconds for Connect to be ready..."
+        sleep 90
+        INPUT_FILE=/opt/docker/connectors/connector_jdbc_customers_template.config OUTPUT_FILE=/opt/docker/connectors/rendered-connectors/connector_jdbc_customers.config /opt/docker/scripts/render-connector-config.sh
+        INPUT_FILE=/opt/docker/connectors/connector_elasticsearch_template.config OUTPUT_FILE=/opt/docker/connectors/rendered-connectors/connector_elasticsearch.config /opt/docker/scripts/render-connector-config.sh
+        curl -s -S -XPOST -H Accept:application/json -H Content-Type:application/json http://connect:8083/connectors/ -d @/opt/docker/connectors/rendered-connectors/connector_elasticsearch.config
+        curl -s -S -XPOST -H Accept:application/json -H Content-Type:application/json http://connect:8083/connectors/ -d @/opt/docker/connectors/rendered-connectors/connector_jdbc_customers.config
```
ybyzek: FYC pull this up to be directly under line 211

rspurgeon:
@ybyzek recent issues addressed.  Thank you for finding validating this.  I validated in a clean directory both the local and docker based demos.  Please let me know if you see any more misbehaviors!

ybyzek (approved):
LGTM, I validated Docker.

PR #699, opened 2020-07-20
DEVX-1832: Update CCloud demos to use ccloud login --save

ybyzek:
Docs change only.  I have validated a local docs build


rspurgeon:
Do you think we should mirror the official docs language to "cover our bases" ?

> Save login credentials or refresh token (in the case of SSO) to local netrc file.

rspurgeon (approved):
LGTM

PR #717, opened 2020-08-03
update ccloud stack image

tikimims:
Updates the ccloud-stack image.


ybyzek (approved):
@tikimims thanks for the PR, LGTM!

PR #728, opened 2020-08-05
DEVX-1989: ccloud local demo: provide specific connector topic configs

rspurgeon:
The configs provided to these topics follow (generally) recommendations in the connector admin docs

### DIFF
--- a/ccloud/start.sh
+++ b/ccloud/start.sh
@@ -7,6 +7,9 @@ source ../utils/ccloud_library.sh
 NAME=`basename "$0"`
 
 echo ====== Verifying prerequisites
+validate_version_confluent_cli_v2 \
+  && print_pass "confluent CLI version ok" \
+  || exit 1
 check_env \
   && print_pass "Confluent Platform installed" \
   || exit 1
@@ -61,7 +64,7 @@ confluent local start zookeeper
 sleep 2
 
 # Start local Kafka with Confluent Metrics Reporter configured for Confluent Cloud
-CONFLUENT_CURRENT=`confluent local current | tail -1`
+CONFLUENT_CURRENT=`confluent local current 2>&1 | tail -1`
 mkdir -p $CONFLUENT_CURRENT/kafka
 KAFKA_CONFIG=$CONFLUENT_CURRENT/kafka/server.properties
 cp $CONFLUENT_HOME/etc/kafka/server.properties $KAFKA_CONFIG
--- a/utils/ccloud_library.sh
+++ b/utils/ccloud_library.sh
@@ -769,11 +769,29 @@ function ccloud::create_acls_connect_topics() {
   serviceAccount=$1
 
   echo "Connect: creating topics and ACLs for service account $serviceAccount"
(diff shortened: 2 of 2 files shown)


utils/ccloud_library.sh
```diff
@@ -769,11 +769,29 @@ function ccloud::create_acls_connect_topics() {
   serviceAccount=$1
 
   echo "Connect: creating topics and ACLs for service account $serviceAccount"
-  for topic in connect-demo-configs connect-demo-offsets connect-demo-statuses _confluent-monitoring _confluent-command ; do
-    ccloud kafka topic create $topic &>/dev/null
-    ccloud kafka acl create --allow --service-account $serviceAccount --operation WRITE --topic $topic --prefix
-    ccloud kafka acl create --allow --service-account $serviceAccount --operation READ --topic $topic --prefix
+
+  TOPIC=connect-demo-configs
+  ccloud kafka topic create $TOPIC --partitions 1 --config "cleanup.policy=compact"
```
ybyzek, line 774: @rspurgeon I understand and agree with the objective here is to configure different partition counts for each of the connect topics.  

1. Can you please elaborate on how partitions were selected and known to be good as 1/6/3 for each respective one?
2. How did we miss this?  Do you think I forgot to validate this version of the demo (perhaps only ran the Docker version) or did some behavior change?

rspurgeon: 1. I followed this guide: https://docs.confluent.io/current/connect/references/allconfigs.html#common-worker-configuration  I did not follow it exactly, however, as CCloud charges per partition, so adding a lot of partitions for topics seems like bad form for exploratory users.  I used instinct for the 6 and 3 respectively, the 1 is required on the configs.
2. I'm not clear.  I'm not sure what other demos exist that run local connect workers to cloud and could possibly have been affected.  Assuming other ones are using the ccloud stack functions, they should be covered by this.

ybyzek: @rspurgeon ack and thanks

ybyzek (approved):
@rspurgeon nice PR, thanks for the quick turnaround.  I haven't validated (assuming you have).  Please note https://github.com/confluentinc/examples/pull/728/files#r465813163 -- should this be removed before merging?  Otherwise, LGTM!

ccloud/start.sh
```diff
@@ -61,7 +64,7 @@ confluent local start zookeeper
 sleep 2
 
 # Start local Kafka with Confluent Metrics Reporter configured for Confluent Cloud
-CONFLUENT_CURRENT=`confluent local current | tail -1`
+CONFLUENT_CURRENT=`confluent local current 2>&1 | tail -1`
```
ybyzek, line 67: @rspurgeon is this a wrong commit from previous PR?

rspurgeon: It's necessary for "late model" `confluent` versions (as of Aug 2020).  It's snuck in here as I missed it for DEVX-1988.  I think it's harmless here but can be removed if you prefer.

ybyzek: Ah, it was missed in previous PR.  👍 include here ;) LGTM

PR #690, opened 2020-06-30
DEVX-1881: microservices-orders: Prepares for usage of latest kafka-s…

rspurgeon:
…treams-examples

Right now this will require running w/ a new variable to test with
a branch of kafka-streams-examples not yet merged.  For example:

`KAFKA_STREAMS_BRANCH=DEVX-1881 ./start.sh`

The docker version of the demo currently uses a manually created kafka-streams-examples image built from the latest source on kafka-streams-examples/DEVX-1881 and hosted on Docker Hub under cnfldemos (https://hub.docker.com/repository/docker/cnfldemos/kafka-streams-examples).

I used the following example workflow to validate against CCloud:

WARN: Create/use a dedicated environment/kafka cluster for this.

* build kafka-streams-examples changes locally
`(cd kafka-streams-examples;git fetch;git checkout DEVX-1881;git pull --rebase;mvn package -DskipTests)`

* delete topics from any previous runs:
 `ccloud kafka topic list -o json | jq -r '.[].name' | xargs -I '{}' ccloud kafka topic delete '{}'`

* pre-create topics:
`TOPICS=(orders orders-enriched payments customers order-validations warehouse-inventory _confluent-monitoring);printf "%s\n" "${TOPICS[@]}" | xargs -I '{}' ccloud kafka topic create '{}'`

* This command will run the microservices
`JAR=$PWD/kafka-streams-examples/target/kafka-streams-examples-5.5.0-standalone.jar BOOTSTRAP_SERVER=pkc-12345.us-central1.cloud.confluent.cloud:9092 SCHEMA_REGISTRY_URL=https://psrc-12345.us-central1.cloud.confluent.cloud RESTPORT=10000 CONFIG_FILE=/Users/username/.ccloud/config-file ./scripts/run-services.sh > ./services.log 2>&1`

* This will read Avro payment data from the payments topic
`kafka-avro-console-consumer --consumer.config /Users/username/.ccloud/config-file --bootstrap-server pkc-12345.us-central1.cloud.confluent.cloud:9092 --property schema.registry.url=https://psrc-12345.us-central1.cloud.confluent.cloud --property schema.registry.basic.auth.user.info=SRKEY:SRSECRET --property basic.auth.credentials.source=USER_INFO --topic payments --from-beginning`

PR #731, opened 2020-08-05
Update OrderDetailsService.java

rocpoc:
A necessary addition for the tests to pass when following the microservices tutorial. If you look at the solutions file, this is included.

Thanks!

### DIFF
--- a/microservices-orders/exercises/OrderDetailsService.java
+++ b/microservices-orders/exercises/OrderDetailsService.java
@@ -192,6 +192,10 @@ public class OrderDetailsService implements Service {
   }
 
   private boolean isValid(final Order order) {
+    final Long customerId = order.getCustomerId();
+    if (customerId == null) {
+      return false;
+    }
     if (order.getQuantity() < 0) {
       return false;
     }


ybyzek:
@rocpoc 

1. What is the error message without this PR?

2. What is the command you are running that generates that error?

3. Looking at the solution at https://github.com/confluentinc/kafka-streams-examples/blob/5.5.1-post/src/main/java/io/confluent/examples/streams/microservices/OrderDetailsService.java#L182-L190, these lines are not included.  So it's not clear what you mean by "If you look at the solutions file, this is included." -- can you please provide a link to what you are looking at?

cc: @rspurgeon

ybyzek:
@rspurgeon note: the lines added into this PR were removed by https://github.com/confluentinc/kafka-streams-examples/pull/346, so could you look into this issue here?  Also, not sure if related or orthogonal, but please also see https://github.com/confluentinc/examples/pull/690#issuecomment-669936810

rspurgeon:
@rocpoc Thanks for the PR.  I removed this code because the return value of `order.getCustomerId()` is a primative `long` and cannot be null, so the expression `customerId == null` can never be true.  There was an implicit cast between `long` and `Long` here.  I validated this could be removed with IDEA's code analyzer. 

Could you let me know what tests you ran where you see a failure?  Thanks

rocpoc:
@rspurgeon thanks for the information!

Looking at my project files, I believe I was working off of the 5.5.0 Post branch. It had been just over a month since I checked on it, but after cloning the repository again, the files now work as intended and don't need any fixes. 

Someone must have fixed the issue in the last month and I didn't see it. Thanks!

Issue #723, opened 2020-08-05
ccloud_library.sh ccloud::create_ccloud_stack throws error on CentOS 7


mkieboom:
The create_ccloud_stack function in the ccloud_library.sh works perfectly fine when running it on MacOS but fails with below error when running on Linux/CentOS 7.

```
Error: accepts 1 arg(s), received 0
Usage:
  ccloud kafka cluster use <id> [flags]

Global Flags:
  -h, --help            Show help for this command.
  -v, --verbose count   Increase verbosity (-v for warn, -vv for info, -vvv for debug, -vvvv for trace).

Kafka cluster id is empty
ERROR: Could not create cluster. Please troubleshoot
```

ybyzek:
`ccloud_library.sh` is currently validated only on macOS, we have not yet validated it on other OS.

I'll create an internal jira to track this issue.  We either need to test on both, otherwise add documentation on the OS test coverage

mkieboom:
Understood, thanks @ybyzek

ybyzek:
@mkieboom thanks for reporting 👍

ybyzek:
@mkieboom @rspurgeon is this GitHub issue resolved by https://github.com/confluentinc/examples/pull/737 ?

rspurgeon:
I believe so, however, I did not personally validate this change on a Linux system, instead verified the code change on MacOS and leaned on @mkieboom to verify on Linux.  We can close this AFAIK unless you'd like me to add to my list a validation on Linux.

Lay3r8:
> `ccloud_library.sh` is currently validated only on macOS, we have not yet validated it on other OS.
> 
> I'll create an internal jira to track this issue. We either need to test on both, otherwise add documentation on the OS test coverage

May I kindly suggest that you guys add a prerequisite section on your docs and tutorials (like this one: https://kafka-tutorials.confluent.io/ccloud-produce-consume/kafka.html) to avoid people trying for hours to do something that is doomed to fail because they don't have the correct OS?

ybyzek:
@Lay3r8 

[...]

PR #737, opened 2020-08-11
Fix for linux support #723

mkieboom:


utils/ccloud_library.sh
```diff
@@ -305,8 +305,7 @@ function ccloud::find_cluster() {
   CLUSTER_REGION=$3
 
   local FOUND_CLUSTER=$(ccloud kafka cluster list -o json | jq -c -r '.[] | select((.name == "'"$CLUSTER_NAME"'") and (.provider == "'"$CLUSTER_CLOUD"'") and (.region == "'"$CLUSTER_REGION"'"))')
-  local FOUND_COUNT=$(echo "$FOUND_CLUSTER" | xargs | wc -l | xargs)
-  [[ $FOUND_COUNT -eq 1 ]] && {
+  [[ ! -z $FOUND_CLUSTER ]] && {
```
rspurgeon: Let's add quoting to avoid issued when FOUND_CLUSTER does not get set

```suggestion
  [[ ! -z "$FOUND_CLUSTER" ]] && {
```

rspurgeon (approved):
LGTM, will merge

PR #748, opened 2020-08-17
fixes for replicator-security

sdandu-gh:
Hit a couple issues in testing replicator-security on 6.0.x (it is possible this needs to be merged to earlier branches).

1) unsecure wasn't working because of a missing indent for services -> srcZookeeper
2) Generally I needed to have the .env symbolic link in the main replicator-security folder in order to run the commands successfully in the README, otherwise they weren't able to find the environment variables (CONFLUENT_DOCKER_TAG, etc..)

### DIFF
--- a/replicator-security/.env
+++ b/replicator-security/.env
@@ -1,0 +1,1 @@
+../utils/config.env
--- a/replicator-security/unsecure/docker-compose.yml
+++ b/replicator-security/unsecure/docker-compose.yml
@@ -13,7 +13,7 @@
 ---
 version: "2.2"
 services:
-srcZookeeper:
+  srcZookeeper:
     image: ${REPOSITORY}/cp-zookeeper:${CONFLUENT_DOCKER_TAG}
     restart: always
     hostname: srcZookeeper


replicator-security/.env
```diff
@@ -0,0 +1 @@
+../utils/config.env
```
ybyzek, line 1: @sdanduConf can you please clarify why this is required in the `replicator-security` folder (versus just the child folders?)

replicator-security/unsecure/docker-compose.yml
```diff
@@ -13,7 +13,7 @@
 ---
 version: "2.2"
 services:
-srcZookeeper:
+  srcZookeeper:
```
ybyzek, line 16: This issue was introduced in 74068fdd0bf5a9a2848f775a289ece4aa2caaac2 (not clear why that change was made @rspurgeon ?)

ybyzek (commented):
@sdanduConf did this run successfully?  AFAICT it should fail because `scripts` isn't accessible to the docker-compose files in the subfolders.  I will fix this (PR available at https://github.com/confluentinc/examples/pull/749) but this issue went unnoticed for a long time, which suggests that neither I nor anyone in the community has run these in a long time.

@thomaskwscott has anyone reported issues?

replicator-security/unsecure/docker-compose.yml
rspurgeon: It's a syntax error, so clearly not intended.  Maybe it was introduced becuase of a merge conflict manual resolution typo and not caught after the merge as we don't alwasy test, or have automated tests, for merge commits.

sdandu-gh:
> @sdanduConf did this run successfully?  AFAICT it should fail because `scripts` isn't accessible to the docker-compose files in the subfolders.  I will fix this (PR available at https://github.com/confluentinc/examples/pull/749) but this issue went unnoticed for a long time, which suggests that neither I nor anyone in the community has run these in a long time.
> 
> 
> 
> @thomaskwscott has anyone reported issues?

Yes, this worked for me when I run from the main replicator-security folder which is why I needed the .env file there as opposed to the subfolders

ybyzek:
@sdanduConf Ah!  Presumably you are running the example as `docker-compose -f unsecure/docker-compose.yml up -d`.  (I think I just forgot how this demo was meant to be run).  This means PR #749 would cause the demo to fail and I should probably cancel that PR.

ybyzek (approved):
Thanks @sdanduConf LGTM

PR #750, opened 2020-08-17
SEC-1307: Backport "log4j replacement with confluent repackaged version"

niteshmor:
This pull request cherry-picks a commit from `master` to add explicit `io.confluent:confluent-log4j` dependency, because we blacklist the (vulnerable) tranistive dependency `log4j:log4j` of `org.slf4j:slf4j-log4j12` explicitly in `common`. `io.confluent:confluent-log4j` is a drop-in replacement of `log4j:log4j` with a fix for the vulnerability.

### DIFF
--- a/clients/avro/pom.xml
+++ b/clients/avro/pom.xml
@@ -73,7 +73,12 @@ http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <dependency>
        <groupId>org.slf4j</groupId>
        <artifactId>slf4j-log4j12</artifactId>
-     </dependency>
+   </dependency>
+   <!-- Use a repackaged version of log4j with security patches. Default log4j v1.2 is a transitive dependency of slf4j-log4j12, but it is excluded in common/pom.xml -->
+    <dependency>
+        <groupId>io.confluent</groupId>
+        <artifactId>confluent-log4j</artifactId>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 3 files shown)


niteshmor:
PR build will pass once the following merges: https://github.com/confluentinc/common/pull/293

andrewegel (approved):

ybyzek (commented):
@niteshmor just to confirm (and for my own education), `confluent-log4j` is intended to be in addition to `slf4j-log4j12`, not instead of?

niteshmor:
> @niteshmor just to confirm (and for my own education), `confluent-log4j` is intended to be in addition to `slf4j-log4j12`, not instead of?

Yes. That's correct. `log4j` is a transitive dependency of `slf4j-log4j12`, and `confluent-log4j` only replaces `log4j`.

ybyzek:
@niteshmor thanks for the explanation.  Also, based on visual diff, it looks like #750 and #751 are identical -- is that right?  Is there are reason there are 2 PRs instead of a single PR to `5.4.x` which would then be pint merged all the way through?

niteshmor:
Reason to do two different PRs is that these changes are already made to `6.0.x` and `master`. `pint merge` doesn't really have a way to specify "start at version branch x but stop at x+n". So it would have required manual cherry-picking anyways.
With two PRs (one for 5.4.x and another for 5.5.x), I can merge the PRs and then run a `pint merge -x ours` to make sure everything is merged well. 

If we didn't have these changes in 6.0.x and master, then yes: a single PR + pint merge would have been sufficient.

PR #764, opened 2020-08-24
DOCS-5157: code modification to work for both SR tutorials

ybyzek:
Code changes to enable extension of SR tutorial https://github.com/confluentinc/docs/pull/5493

### DIFF
--- a/clients/avro/src/main/java/io/confluent/examples/clients/basicavro/ConsumerExample.java
+++ b/clients/avro/src/main/java/io/confluent/examples/clients/basicavro/ConsumerExample.java
@@ -25,16 +25,18 @@ public class ConsumerExample {
     @SuppressWarnings("InfiniteLoopStatement")
     public static void main(final String[] args) throws IOException {
 
-        if (args.length != 1) {
-          System.out.println("Please provide command line arguments for configPath");
-          System.exit(1);
+        Properties props = new Properties();
+        if (args.length < 1) {
+          // Backwards compatibility, assume localhost
+          props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9092");
+          props.put(AbstractKafkaSchemaSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, "http://localhost:8081");
+        } else {
+          // Load properties from a local configuration file
+          // Create the configuration file (e.g. at '$HOME/.confluent/java.config') with configuration parameters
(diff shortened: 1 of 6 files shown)


clients/avro/src/main/java/io/confluent/examples/clients/basicavro/ConsumerExample.java
```diff
@@ -44,4 +60,15 @@ public static void main(final String[] args) {
...
+    public static Properties loadConfig(final String configFile) throws IOException {
+      if (!Files.exists(Paths.get(configFile))) {
+        throw new IOException(configFile + " not found.");
+      }
+      final Properties cfg = new Properties();
+      try (InputStream inputStream = new FileInputStream(configFile)) {
+        cfg.load(inputStream);
+      }
+      return cfg;
+    }
```
rspurgeon: As they are identical, I recommend we share this function across the two examples.  Since they are public and static, would be a simple matter of deleting one of them and adding an include for the other one's function.  LMK if you want me to make the change and push or I can review after you make it.

ybyzek: @rspurgeon if you have the availability, it would be fantastic if you could verify & push the change.  If not, please let me know.

londoncalling (approved):
LGTM!

clients/avro/src/main/java/io/confluent/examples/clients/basicavro/ConsumerExample.java
ybyzek: Thinking about this more, while this suggestion achieves deduplication, the downside is:

1. It creates an unnecessary dependency between the two classes
2. It's not a logical dependency
3. Anyone copying one of the examples will encounter friction with this dependency

I have an idea for modifying this and removing the function completely.  WDYT about changes in latest commit https://github.com/confluentinc/examples/pull/764/commits/8ca1dfe74b6f9f5db36ddfc09dc083cfb941cfc4 ?

rspurgeon: Because it's compiled within the same project in the same namespace, IMO 1 & 2 aren't concerns. You just need a place to put a function, Ideally there would be a little file utility class to put it, but old school Java makes that annoying.  For 3. if the point is for someone to copy and paste the example from the docs or from this example then I agree make it easier for that.  Regardless, for this simple example I have no objection to in-lining the code or leaving the duplicate function.

rspurgeon (approved):
LGTM

PR #777, opened 2020-09-08
Fix comparison operator typo in ccloud_library.sh

gitfrog0:
Normally, the correct credentials are found with only the resource_id, but I was trying to repurpose an existing cloud environment and things didn't work right due to the comparison being ignored.


ybyzek:
cc @rspurgeon for review

rspurgeon (approved):
Nice find @gitfrog0 thank you for the PR.  I'll merge this through

PR #792, opened 2020-09-17
DEVX-2086: Update client cloud examples based on recommendations list…

ybyzek:
…ed in our docs (for librdkafka-based clients

### DIFF
--- a/clients/docs/includes/configs/cloud/librdkafka-sr.config
+++ b/clients/docs/includes/configs/cloud/librdkafka-sr.config
@@ -5,6 +5,9 @@ sasl.mechanisms=PLAIN
 sasl.username={{ CLUSTER_API_KEY }}
 sasl.password={{ CLUSTER_API_SECRET }}
 
+# Tuning recommendations
+retries=2147483647
+
 # Confluent Cloud Schema Registry
 schema.registry.url=https://{{ SR_ENDPOINT }}
 basic.auth.credentials.source=USER_INFO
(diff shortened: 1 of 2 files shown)


vdesabou:
This commit will add the `retries` parameter in this [link](https://docs.confluent.io/current/tutorials/examples/clients/docs/python.html) for example, right ?

As a user, when I land on this [page](https://github.com/confluentinc/examples/tree/5.5.1-post/clients/cloud/python), I go directly into producer.py and see [that](https://github.com/confluentinc/examples/blob/4f7f2b7b1bd9786ba84a5236082dc0030be46d3a/clients/cloud/python/producer.py#L38):

```
    # Create Producer instance
    producer = Producer({
        'bootstrap.servers': conf['bootstrap.servers'],
        'sasl.mechanisms': conf['sasl.mechanisms'],
        'security.protocol': conf['security.protocol'],
        'sasl.username': conf['sasl.username'],
        'sasl.password': conf['sasl.password'],
    })
```

I would have expected to see the `retries` in there ? (sorry I'm not a python user :) )

ybyzek:
@vdesabou yes, https://docs.confluent.io/current/tutorials/examples/clients/docs/python.html --> `Setup` section step 3 displays this file as-is, so that page will be updated.

The code example itself will not be updated by this PR mainly because the code is meant to be flexible -- run to CCloud or on-prem or localhost, etc...  Therefore, code can't set `retries` to an explicit value.  Code could set `retries` to the value in the file it reads from (e.g. `'retries': conf['retries'],`) but it's not a requirement for it to be defined so that leaves the problem that if a user doesn't define `retries`, code would error out.  So it's not a clear path how to add to the code itself.  My leaning is to not make code modifications.  WDYT?

ybyzek:
Closing this PR because https://github.com/edenhill/librdkafka/pull/3076 will have this out to users in the mid-October release.  See DEVX jira for more details

PR #800, opened 2020-09-22
DEVX-2113: Remove ZK from AK CLI commands and component configurations

ybyzek:


multi-datacenter/read-topics.sh
```diff
@@ -7,7 +7,7 @@ source ../utils/config.env
 
 echo -e "\n-----dc1-----"
 echo -e "\nlist topics:"
-docker-compose exec broker-dc1 kafka-topics --list --zookeeper zookeeper-dc1:2181
+docker-compose exec broker-dc2 kafka-topics --list --bootstrap-server broker-dc1:29091
```
rspurgeon: was the `dc2` change here intentional?

ybyzek: Great catch, not intentional.  Fixing

rspurgeon (approved):
LGTM

PR #803, opened 2020-09-23
[DEVX-2013] Update rep factor for balancer and audit log topics

sdandu-gh:
With new internal topics in 6.0, if the default is not set, topic creation will fail. This will not impact functionality of any of these demos, but will result in extra errors in the logs and will result in those topics not being available if someone was adding on top of the demo.

Verified all modified demos with updated 6.0 with these new configs.


ybyzek (approved):
@sdanduConf Thank you!!  LGTM

PR #805, opened 2020-09-25
DEVX-2123: Delay reading ksqlDB topics during validation until available

rspurgeon:


rspurgeon:
This is for docker only, I can delay this and verify local & cloud versions as well, or we can do them as independent PRs if necessary

microservices-orders/read-topics-docker.sh
```diff
@@ -33,6 +33,10 @@ docker-compose exec connect kafka-console-consumer --bootstrap-server broker:909
 echo -e "\n-----platinum-----"
 docker-compose exec connect kafka-avro-console-consumer --bootstrap-server broker:9092 --topic platinum --property print.key=true --property key.deserializer=org.apache.kafka.common.serialization.StringDeserializer --property schema.registry.url=http://schema-registry:8081 --from-beginning --max-messages 3 --timeout-ms 10000
 
+for i in {1..10}; do
```
ybyzek: @rspurgeon 

1. If the required check is the *presence* of a topic, can we leverage the function `check_topic_exists` at https://github.com/confluentinc/examples/blob/5.5.1-post/utils/helper.sh#L406-L413?  Example usage: https://github.com/confluentinc/examples/blob/5.5.1-post/ccloud/start-docker.sh#L80-L83

2. If the required check is that the topic has at least one message, then can you convert this into a function in the helper.sh file and use the retry method as used in other scripts?

rspurgeon: The library dependency was not already present so I decided not to introduce it, considering the ease of the addition and it's all built in bash I don't see the dependency as necessary (IMHO).  It's really just a single liner which I've broken into 3 lines for readability

ybyzek: IMO it's better to reuse the function, even at the cost of pulling in the library.  The value being:

1. There are differences in behavior even for "built in bash" functions
2. Promote consistency in logical handling of this check
3. Don't need to rethink the bash.  This differs already in the check, output, sleep, etc

rspurgeon:
> This is for docker only, I can delay this and verify local & cloud versions as well, or we can do them as independent PRs if necessary

This delay does not appear necessary for local version of the demonstration

rspurgeon:
ccloud version of the demo also functions without any delay

rspurgeon:
@ybyzek incorporated helper.sh, please review

ybyzek (approved):
@rspurgeon LGTM, thanks!

PR #813, opened 2020-09-28
DEVX-2070: Update default datagen version in Dockerfile-local to 6.0.0

rspurgeon:
Updating the default only, this has no material affect on the current demo behavior as it is set in script by start-docker.sh

### DIFF
--- a/ccloud/Dockerfile-local
+++ b/ccloud/Dockerfile-local
@@ -17,7 +17,7 @@ ARG CP_VERSION
 
 FROM confluentinc/cp-server-connect-base:$CP_VERSION
 
-ARG CONNECTOR_VERSION=6.0.0-SNAPSHOT
+ARG CONNECTOR_VERSION=6.0.0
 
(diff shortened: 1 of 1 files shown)


ccloud/Dockerfile-local
```diff
@@ -17,7 +17,7 @@ ARG CP_VERSION
 
 FROM confluentinc/cp-server-connect-base:$CP_VERSION
 
-ARG CONNECTOR_VERSION=5.5.0-SNAPSHOT
+ARG CONNECTOR_VERSION=6.0.0-SNAPSHOT
```
ybyzek: @rspurgeon should this be `6.0.0` instead?

rspurgeon: @ybyzek .. yes.  Thank you! 🤦

rspurgeon: @ybyzek did you see my correction for this ?

ybyzek (approved):
LGTM

PR #815, opened 2020-09-30
Cleanup Streams init

simPod:
Remove handling of "Streams never finished rebalancing on startup" as it's no longer required (I believe it is some relict to prevent a bug that existed in Kafka v0.x)

### DIFF
--- a/microservices-orders/exercises/EmailService.java
+++ b/microservices-orders/exercises/EmailService.java
@@ -21,9 +21,6 @@ import org.apache.kafka.streams.kstream.Consumed;
 import java.io.IOException;
 import java.util.Optional;
 import java.util.Properties;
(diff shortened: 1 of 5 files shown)


ybyzek (changes requested):
@simPod these changes cannot be made independently from the original source code at https://github.com/confluentinc/kafka-streams-examples/tree/6.0.0-post/src/main/java/io/confluent/examples/streams/microservices .  The files here are from that `kafka-streams-examples` repo, with certain lines removed.  So this PR could only be approved if the corresponding changes are first merged in `kafka-streams-examples`

simPod:
https://github.com/confluentinc/kafka-streams-examples/pull/364

PR #816, opened 2020-09-30
REP-1233 fix for slow start up due to cpus restriction

thomaskwscott:
CPU limits were causing brokers to start too slowly for the embedded HTTP server. This Pr removes these restriction. 

All of these demos have been run locally.


ybyzek (approved):
@thomaskwscott great sleuthing, LGTM.  This also solves the mystery why the multi-datacenter demo didn't exhibit the same behavior—that demo has no CPU restrictions (https://github.com/confluentinc/examples/blob/6.0.0-post/multi-datacenter/docker-compose.yml)

PR #819, opened 2020-10-01
DEVX-2134: Add step and integrate verbiage from prerequisite

tikimims:
Added step and integrate verbiage from a prerequisite.


tikimims:
@ybyzek Here's what I did here:

- Removed the prerequisite involving the `ccloud login --save` command
- Added a new step and integrated the removed prerequisite
- Ran a full build and its clean

What do you think about this approach? If it looks okay, can you merge as well?

Thanks so much!

ybyzek (approved):
LGTM @tikimims 🥇

PR #820, opened 2020-10-02
Tweak Cloud Java client configs for correctness and improved comments

ijuma:
* Add `acks=all` to avoid data loss during kafka rolls (librdkafka defaults to
`acks=all` since 1.0).
* Add `client.dns.lookup=use_all_dns_ips` for older clients (it's the
default since Apache Kafka 2.6)


clients/docs/includes/configs/cloud/java-sr.config
```diff
@@ -1,10 +1,13 @@
-# Kafka
+# Kafka client (producer/consumer/admin)
 bootstrap.servers={{ BROKER_ENDPOINT }}
 security.protocol=SASL_SSL
 sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required username="{{ CLUSTER_API_KEY }}" password="{{ CLUSTER_API_SECRET }}";
-ssl.endpoint.identification.algorithm=https
 sasl.mechanism=PLAIN
 
+# Kafka producer
```
ybyzek: @ijuma (cc: @rspurgeon ): a few questions about adding these settings into these files

1. Are these settings advised for Confluent Cloud only, or also on-prem?  (There are additional files to modify for on-prem)
2. If these settings are advised for any deployment (depends on answer above), then why aren't they made in the AK producer code base itself?
3. These files are also used in other contexts, e.g. as ingestion to CP components (example: https://docs.confluent.io/current/tutorials/examples/clients/docs/rest-proxy.html#setup).  Do they apply there too?  If not, maybe this isn't the right file to add?
4. Stepping back, given that these are optimizations and not required, do they even belong in this file?  Original intention was that these were just minimum set of required parameters to _connect_ to a Kafka cluster.  Should this PR go just to docs?

clients/docs/includes/configs/cloud/java.config
```diff
@@ -1,6 +1,9 @@
-# Kafka
+# Kafka client (producer/consumer/admin)
```
ybyzek: If we do decide to keep the `Kafka producer` section (pending other comment), perhaps we can rename the headings as:

`# Kafka producer, consumer, and admin`
`# Kafka producer only`

clients/docs/includes/configs/cloud/java-sr.config
```diff
@@ -1,10 +1,13 @@
-# Kafka
+# Kafka client (producer/consumer/admin)
 bootstrap.servers={{ BROKER_ENDPOINT }}
 security.protocol=SASL_SSL
 sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required username="{{ CLUSTER_API_KEY }}" password="{{ CLUSTER_API_SECRET }}";
```
ybyzek: We will need to test this PR to validate removal of `ssl.endpoint.identification.algorithm`, just to double-check.

There are also many references in docs with this configuration parameter.  Do those need to be modified as well?

```
cp-component/connect-cloud-config.rst:        ssl.endpoint.identification.algorithm=https
cp-component/connect-cloud-config.rst:        consumer.ssl.endpoint.identification.algorithm=https
cp-component/connect-cloud-config.rst:        producer.ssl.endpoint.identification.algorithm=https
cp-component/connect-cloud-config.rst:         database.history.consumer.ssl.endpoint.identification.algorithm=https
cp-component/connect-cloud-config.rst:         database.history.producer.ssl.endpoint.identification.algorithm=https
cp-component/connect-cloud-config.rst:        ssl.endpoint.identification.algorithm=https
cp-component/connect-cloud-config.rst:        consumer.ssl.endpoint.identification.algorithm=https
cp-component/connect-cloud-config.rst:        producer.ssl.endpoint.identification.algorithm=https
cp-component/connect-cloud-config.rst:            "database.history.consumer.ssl.endpoint.identification.algorithm": "https",
cp-component/connect-cloud-config.rst:            "database.history.producer.ssl.endpoint.identification.algorithm": "https",
cp-component/c3-cloud-config.rst:        confluent.controlcenter.streams.ssl.endpoint.identification.algorithm=https
cp-component/c3-cloud-config.rst:        confluent.monitoring.interceptor.ssl.endpoint.identification.algorithm=https
migrate-topics-on-cloud-clusters.rst:    ssl.endpoint.identification.algorithm=https
migrate-topics-on-cloud-clusters.rst:    ssl.endpoint.identification.algorithm=https
migrate-topics-on-cloud-clusters.rst:        "src.kafka.ssl.endpoint.identification.algorithm":"https",
migrate-topics-on-cloud-clusters.rst:        "dest.kafka.ssl.endpoint.identification.algorithm":"https",
migrate-topics-on-cloud-clusters.rst:  ssl.endpoint.identification.algorithm=https
migrate-topics-on-cloud-clusters.rst:  producer.ssl.endpoint.identification.algorithm=https
migrate-topics-on-cloud-clusters.rst:   ssl.endpoint.identification.algorithm=https
migrate-topics-on-cloud-clusters.rst:     ssl.endpoint.identification.algorithm=https
```

clients/docs/includes/configs/cloud/java-sr.config
ijuma: The intent is for these to be the default for CCloud. Some of them will become the default in AK 3.0, but that will take a while to reach users. `acks=all` is for correctness and `linger.ms=5` is important for reasonable efficiency (it affects us directly in MT clusters for example).

clients/docs/includes/configs/cloud/java.config
ijuma: Sounds good, I'll update once we reach some consensus on the other discussion.

clients/docs/includes/configs/cloud/java-sr.config
ijuma: Possibly, but I'd prefer the teams owning those areas to verify. Note that the setting is redundant if you're using a client released within the last 2+ years. It doesn't harm. Also, for CCloud, it's always been optional.

ybyzek: > Note that the setting is redundant if you're using a client released within the last 2+ years. It doesn't harm. 

@ijuma if this has been unnecessary for the last 2+ years, that is good enough for me.  (In contrast, if it were a new option introduced in 2.6, we would probably not remove it for a while).

Note: we will still want to validate so please allow us some time to do so.

ijuma: Totally agreed. That's why I added `client.dns.lookup=use_all_dns_ips` even though it's the default in 2.6.

clients/docs/includes/configs/cloud/java.config
ybyzek: Brainstorming out loud: maybe we tweak the headings to address the point "Stepping back, given that these are optimizations and not required, do they even belong in this file? Original intention was that these were just minimum set of required parameters to connect to a Kafka cluster. Should this PR go just to docs?"

For example, first heading could be:

`# Required connection settings for Kafka producer, consumer, and admin`

Second heading could be:

`# Recommended settings for optimal Kafka producer performance`

That would more clearly communicate the differences.  Not just that they are for producer, but that they are optional (but recommended)...versus **required** for connectivity.

WDTY @ijuma @rspurgeon  ?

clients/docs/includes/configs/cloud/java-sr.config
ybyzek: @ijuma apologies if I'm just missing it, but I don't see `client.dns.lookup=use_all_dns_ips` in the PR yet?

clients/docs/includes/configs/cloud/java-sr.config
```diff
@@ -1,10 +1,13 @@
-# Kafka
+# Kafka client (producer/consumer/admin)
```
junrao: reconnect.backoff.max.ms defaults to 1 sec in AK. In CCloud, we have seen frequent reconnects due to bad user/passwd, which increases the load on the broker. I wonder if we should change the default to sth larger.

ijuma: Sounds good. What would you suggest? 5 seconds? Also, maybe we want to change the AK default as part of https://cwiki.apache.org/confluence/display/KAFKA/KIP-580%3A+Exponential+Backoff+for+Kafka+Clients

clients/docs/includes/configs/cloud/java-sr.config
ijuma: Good point, I somehow lost this change when I moved the PR from the other repo to this one. Fixed.

clients/docs/includes/configs/cloud/java.config
ijuma: This idea seems promising. What about `acks=all` though? That is required for not losing data, it's not about optimal performance. Where would you have that?

clients/docs/includes/configs/cloud/java-sr.config
junrao: Since this should be rare, maybe even setting it to 30 secs. Yes, it would be useful to bump up the defaults in AK too.

ybyzek:
@ijuma just as FYI, here are 3 additional PRs to remove `ssl.endpoint.identification.algorithm` from all Integration Architecture team examples:

1. https://github.com/confluentinc/examples/pull/821
2. https://github.com/confluentinc/cp-demo/pull/256
3. https://github.com/confluentinc/cp-all-in-one/pull/40

clients/docs/includes/configs/cloud/java-sr.config
```diff
@@ -1,9 +1,13 @@
...
 bootstrap.servers={{ BROKER_ENDPOINT }}
 security.protocol=SASL_SSL
 sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required username="{{ CLUSTER_API_KEY }}" password="{{ CLUSTER_API_SECRET }}";
-ssl.endpoint.identification.algorithm=https
 sasl.mechanism=PLAIN
+client.dns.lookup=use_all_dns_ips
```
ybyzek, line 7: `client.dns.lookup=use_all_dns_ips` is an optimization, it is not required, and so for reasons already mentioned in this PR, maybe this shouldn't be here either and rather moved to `# Recommended settings for optimal Kafka producer performance`...or removed completely along with other optimizations and just. moved into docs.

ijuma: This is not an optimization, it is required for correctness.

ybyzek: Going back to https://github.com/confluentinc/examples/pull/820#discussion_r498894181:

> Original intention was that these were just minimum set of required parameters to _connect_ to a Kafka cluster. 

To connect: the only required configurations are the following 4 configuration parameters:

```
bootstrap.servers={{ BROKER_ENDPOINT }}
security.protocol=SASL_SSL
sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required username="{{ CLUSTER_API_KEY }}" password="{{ CLUSTER_API_SECRET }}";
sasl.mechanism=PLAIN
```

Since clients can still work without setting `client.dns.lookup=use_all_dns_ips`, it is not required in these configuration files (let's not bikeshed about the word `optimization` versus `correctness`).

My leaning is still

1. Keep the configuration files in this repo to the original 4 required parameters (`bootstrap.servers`, `security.protocol`, `sasl.jaas.config`, `sasl.mechanism`), to focus the user on those parameters absolutely required to connect.
2. For the rest of the recommendations (`acks`, `linger.ms`, `reconnect.backoff.max.ms`, `client.dns.lookup`, etc) make sure they are added to Confluent documentation, e.g. https://docs.confluent.io/current/cloud/using/config-client.html#recommendations.   These parameters are good additions but it may confuse users if they are in these files.  Confusing because: they are not required for connectivity, some apply just for the producer but not consumer or admin, and they require deeper explanation.

ijuma: @ybyzek I did not intend to bikeshed, I think it's an important distinction. I have been in several incidents due to `use_all_dns_ips` not being set. @Evelyn-Bayes asked a related question a few days ago:

https://confluent.slack.com/archives/CF6E4C48N/p1601611374088300

I am a big fan of keeping configs as simple as we can, but I do think that having unsafe configs by default hurts our users in a big way and it's extremely costly for us. The sad truth is that many of our Cloud users don't read the documentation until it's too late. Often only after an incident that involves GTS, Engineering and causes a serious outage on their end.

I think a case can be made that `linger.ms=5` should be be removed from the PR (even though the inefficiency results in a worse experience to our users) and I can go along with that, but the rest are really essential unless you're an expert user and you know what you're doing.

ybyzek: 💯 agree that these additional configuration parameters help users have a better experience.  (I did not mean to convey that these additional configuration parameters are not important.)

We should strive to help users in three ways, in order of highest-to-lowest impact:

- Enhance Apache Kafka by changing the defaults, like `client.dns.lookup=use_all_dns_ips` which became the default with https://issues.apache.org/jira/browse/KAFKA-9313, a jira that I filed last year
- In-product help (Confluent Cloud UI in this case)
- Out-of-product documentation that lives somewhere else (like docs.confluent.io)

The crux of this PR is how to address #2.  Your point makes sense, so let's find a way to present the information in the Confluent Cloud UI, it will have the broadest reach and greatest potential to get used so that users configure their clients for better behavior. 👍 

In doing so, we have to address a concern : users copy/paste things exactly as shown to them, and don't do their own research to understand what the configuration parameters mean ("The sad truth is that many of our Cloud users don't read the documentation until it's too late." --> agree).  If the UI shows them X, they will blindly copy and paste X to EVERYTHING: producers, consumers, CLI, etc,...and then they will copy/paste the whole thing again when they want to reuse the properties file to connect to some other cluster...  This causes potential friction points: 

(a) differentiating what is mandatory to connect to Confluent Cloud from what gives better experience
(b) differentiating which configs are for all clients versus just producer versus just consumer, etc
(c) portability of properties file -- users may take the properties intended for CCloud and then blindly reuse the same configuration file for testing on a local cluster.

So in whatever way we present these configurations in the Confluent Cloud UI, it needs to be very clear and crisp, explaining the purpose of each configuration parameter, or each set of configuration parameters.  I might have argued for different blocks of code with separate copy/paste sections, but that involves more investment on the Confluent Cloud UI side, so the lowest cost is putting everything into the same block as done here.  Let's keep iterating and polishing the ideas we had with the comments in this GH issue.  

@ijuma  tl;dr I agree, let's proceed **Would you be able to make another pass at the changes in this PR and embed more elaborate comments (comments into the files themselves, not GH comments) that explain each parameter in more detail, and that address (a), (b), and (c) above?**

Additionally:

1. **AK Defaults**: it sounds like there is some further possibility for improvement with setting AK defaults for `reconnect.backoff.max.ms ` -- is there an Apache Kafka jira for this yet?

2. @Evelyn-Bayes 's message also mentioned `networkaddress.cache.ttl` and `networkaddress.cache.negative.ttl`, that doesn't fit this model...where does that go into the Confluent Cloud UI?

3. **Documentation enhancement** :  https://github.com/confluentinc/docs/pull/4730/files did not update with all the parameters (including) `reconnect.backoff.max.ms`, and the docs at https://docs.confluent.io/current/cloud/using/config-client.html#recommendations may still need to be updated.  Are you going to fix that?

4. Are you going to file a DOCS jira for all the other components to remove `ssl.endpoint.identification.algorithm=https` and add these recommended configurations?

5. FYI I filed 3 PRs to strip `ssl.endpoint.identification.algorithm=https` from all IA demos since it has been the default for 2+ years.

https://github.com/confluentinc/examples/pull/821
https://github.com/confluentinc/cp-demo/pull/256
https://github.com/confluentinc/cp-all-in-one/pull/40

6. But of course if the recommendations that come via this PR Include additional parameters, we will have to add them everywhere as well in the examples (not just for CCloud UI).

ijuma: Totally agreed with the way you described the 3 ways we should help users. I am hoping that we can make `acks=all` and `enable.idempotence=true` the default in Apache Kafka 3.0 for better out of the box behavior in that respect too (@d8tltanc is looking into that in Q4).

Yes, I can add descriptions with the configs. Answers to your numbered questions:

1. There isn't a JIRA, but there is a KIP for introducing exponential backoff for retries. Maybe we can consider making the change as part of that. I pinged the relevant people.
2. Not sure. Also, the defaults are actually reasonable unless a security manager is used. https://github.com/openjdk/jdk/blob/6bab0f539fba8fb441697846347597b4a0ade428/src/java.base/share/classes/sun/net/InetAddressCachePolicy.java#L48. So, maybe it's OK for it to be in docs only.
3. Yeah, I submitted https://github.com/confluentinc/docs/pull/5845 last week and I was planning to add Jun's suggestion there.
4. After we reach consensus and we merge this PR, sure.
5. Great! Will review.
6. You mean all ccloud examples or all examples?

ijuma: @ybyzek I removed linger.ms and tweaked the comments a bit. I think it's best to keep it simple instead of trying to explain in a lot of detail. Please let me know what you think.

clients/docs/includes/configs/cloud/java-sr.config
```diff
@@ -1,10 +1,14 @@
-# Kafka
+# Required connection configs for Kafka producer, consumer, and admin
 bootstrap.servers={{ BROKER_ENDPOINT }}
 security.protocol=SASL_SSL
 sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required username='{{ CLUSTER_API_KEY }}' password='{{ CLUSTER_API_SECRET }}';
 sasl.mechanism=PLAIN
+client.dns.lookup=use_all_dns_ips
```
ybyzek, line 7:
```suggestion

# Best Practice for all Java clients
client.dns.lookup=use_all_dns_ips
```

clients/docs/includes/configs/cloud/java-sr.config
```diff
@@ -1,10 +1,14 @@
...
 
-# Confluent Cloud Schema Registry
+# Recommended Kafka producer config to prevent data loss
```
ybyzek:
```suggestion
# Best practice for Kafka producer to prevent data loss 
```

clients/docs/includes/configs/cloud/java.config
```diff
@@ -1,5 +1,9 @@
-# Kafka
+# Required connection configs for Kafka producer, consumer, and admin
 bootstrap.servers={{ BROKER_ENDPOINT }}
 security.protocol=SASL_SSL
 sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required username='{{ CLUSTER_API_KEY }}' password='{{ CLUSTER_API_SECRET }}';
 sasl.mechanism=PLAIN
+client.dns.lookup=use_all_dns_ips
```
ybyzek, line 7:
```suggestion

# Best Practice for all Java clients
client.dns.lookup=use_all_dns_ips
```

clients/docs/includes/configs/cloud/java.config
```diff
@@ -1,5 +1,9 @@
...
+
+# Recommended Kafka producer config to prevent data loss
```
ybyzek:
```suggestion
# Best practice for Kafka producer to prevent data loss 
```

ybyzek (commented):
Thanks @ijuma .  I left some comments for the existing parameters.  Additionally, is this PR going to address either `reconnect.backoff.max.ms`, `networkaddress.cache.ttl`, or `networkaddress.cache.negative.ttl`, or is that left for a different time?

clients/docs/includes/configs/cloud/java-sr.config
ijuma: Is this really worth it? I think most users won't care about the distinction. I'd just leave it as part of the previous block and remove it in a few months once 2.6+ is a common choice for new apps.

ijuma:
Thanks for the review @ybyzek. I applied some of your suggestions and left a question for the others. With regards to `reconnect.backoff.max.ms`, `networkaddress.cache.ttl` and `networkaddress.cache.negative.ttl`, they will be handled via documentation only. @jolshan is working on improving those.

clients/docs/includes/configs/cloud/java-sr.config
ybyzek: > Is this really worth it? I think most users won't care about the distinction. I'd just leave it as part of the previous block and remove it in a few months once 2.6+ is a common choice for new apps.

Yes, IMO the distinction is important, see (a) from earlier comment:

```
(a) differentiating what is mandatory to connect to Confluent Cloud from what gives better experience
(b) differentiating which configs are for all clients versus just producer versus just consumer, etc
(c) portability of properties file -- users may take the properties intended for CCloud and then blindly reuse the same configuration file for testing on a local cluster.
```

ijuma: The question is why is this important in this case. This config is necessary for correctness, it's the new default, it's safe to use in Cloud and CP, it's safe to use in every client, there is no downside to using it and it's a connection related config. I don't think we're really helping people by separating it (unlike the acks config). If you really insist, I'll make the change you suggested, but I strongly believe it makes it worse rather than better.

ybyzek: @ijuma 

1. If `client.dns.lookup=use_all_dns_ips` is going to be added into these files, then I strongly lean towards having it commented as suggested.

2. wrt "it's safe to use in every client" -- this PR currently modifies only the Java files.  If it's applicable to use in every client, then for consistency this PR should modify the other files as well (see all files within subdirectory https://github.com/confluentinc/examples/tree/6.0.0-post/clients/docs/includes/configs)

ijuma: I mean in every Java client. It's not needed by other clients. The main reason why I'm adding it is that not having this causes outages in CCloud production. It's not really "best practice", it's required.

ybyzek: If the concern is to prevent/avoid outages in CCloud production, then the target user is those running _older_ code, which means they are likely not referring to these pages anyway.  Is it possible that the users who would refer to these pages to get started are likely running AK 2.6 already?

ybyzek (approved):
@ijuma LGTM.

PR #824, opened 2020-10-06
DEVX-2153: Fix ksql-datagen example for KSQL-3812; DEVX-2151: missed …

ybyzek:
…instances

This PR fixes two things:

1. DEVX-2153: remove error in ksql-datagen by removing line `KSQL_LOG4J_OPTS: "-Dlog4j.configuration=file:/etc/ksql/log4j-rolling.properties"` (not supported for ksql-datagen)
2. Missed instances for DEVX-2151: removing `ssl.endpoint.identification.algorithm=https` which is the default

I have validated the ksql-datagen example, and previous PR validated removal of `ssl.endpoint.identification.algorithm=https`


rspurgeon (approved):
LGTM

PR #827, opened 2020-10-06
Fix build issues for migration

joel-hamill:


ybyzek (approved):
@joel-hamill approved.  For clarification, can you please elaborate on the motivation -- are all `:ref:` directives going away?

joel-hamill:
`:ref:` are going away for any content that is *not* in the docs-platform repo (on-premises). `:ref:` links only work if all content is built from the same repo.

PR #830, opened 2020-10-07
DEVX-1778: replace deprecated methods AbstractKafkaAvroSerDeConfig and argument to poll()

awalther28:
Update references to deprecated serde to non-deprecated counter part. Minor change to consumer poll to use Java Duration rather than long in multi-datacenter example. 

Testing:
- ran connect-streams-pipeline along and clients/cloud/java/.../StreamsAvroExample against local CP 
- ran multi-datacenter against docker-compose CP
ran demos prior to changes and after, same behavior


ybyzek (approved):
@awalther28 LGTM, thank you.  I also validated with `mvn clean compile`.

rspurgeon (approved):
LGTM

PR #832, opened 2020-10-08
DEVX-2164: Help with ccloud-stack adoption by showing the CLI commands that are executed

ybyzek:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2164

- show precisely which ccloud commands are run
- give indication of what is free and what incurs charges

### Author Validation

Waiting for docs migration to complete to verify local build of this PR

- [x] Documentation

### Reviewer Tasks

- [x] Documentation

### DIFF
--- a/ccloud/docs/ccloud-stack.rst
+++ b/ccloud/docs/ccloud-stack.rst
@@ -30,9 +30,12 @@ Caution
 =======
 
 This utility uses real |ccloud| resources.
+If you just run ``ccloud-stack`` without enabling |ccloud| ksqlDB and do not create any topics or other resources, then there is no billing charge.
+If you run ``ccloud-stack`` with enabling |ccloud| ksqlDB, or create topics, run a |ccloud| connector, create ksqlDB queries, or enable other resources, then there will be a billing charge.
 To avoid unexpected charges, carefully evaluate the cost of resources before launching the utility and ensure all resources are destroyed after you are done running it.
 
-To help evaluate the cost of running ``ccloud-stack``, here is a list of |ccloud| CLI commands issued by the utility that create resources in |ccloud| (function ``ccloud::create_ccloud_stack()`` source code is in :devx-examples:`ccloud_library|utils/ccloud_library.sh`). By default, the ksqlDB app is not created with ``ccloud-stack``, you have to explicitly set it.
+Here is a list of |ccloud| CLI commands issued by the utility that create resources in |ccloud| (function ``ccloud::create_ccloud_stack()`` source code is in :devx-examples:`ccloud_library|utils/ccloud_library.sh`).
+By default, the |ccloud| ksqlDB app is not created with ``ccloud-stack``, you have to explicitly enable it.
 
 .. code-block:: text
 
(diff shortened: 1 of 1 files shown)


ybyzek:
@caleb-grillo @bechols -- can you please review the PR to ensure the statements about cost/billing are valid?

ccloud/docs/ccloud-stack.rst
```diff
@@ -30,8 +30,32 @@ Caution
 =======
 
 This utility uses real |ccloud| resources.
+If you just run ``ccloud-stack`` without enabling |ccloud| ksqlDB and do not create any topics or other resources, then there is no billing charge.
```
rspurgeon: IMO we remove this.  We don't want users to use the utility unless it's for something useful and we don't want them to create resources (even logical non-cost associated ones) for no reason.

ybyzek: There is still immense value in running ``ccloud-stack`` without ksqlDB because it properly creates the credentials for each resource (and wildcard ACLs).  This addresses some user friction we have observed from new users.  The key with this PR was to be transparent about what the cost is, to encourage users to use ``ccloud-stack``.  It's a value add to say it's free, and the outcome is proper credentials.

rspurgeon: >and do not create any topics or other resources, then there is no billing charge.

Can you clarify this portion of the sentence as maybe I'm not understanding what you're implying with this?

rspurgeon: If you're implying that users should/could use ccloud-stack to create environments and credentials (non-billable) for _later_ use in contexts that _are_ billable, I think the message is potentially confusing.  Nothing about a new enviornment has value until a user creates a topic pushes data, which comes with a cost.  I think you could reword this to better convey what I think you're trying to say which is, you can use this to "setup" an environment with proper credentials for use later.  Regardless, the "later use" will result in a charge.

ybyzek: @rspurgeon what do you think about the revised version?

rspurgeon: Thanks, it's clearer for sure IMO.  WDYT about something like (I think I'm interpreting you correctly with this):

> ccloud-stack can be useful to "setup" an environment for other demos or use cases you may require by creating an environment and credentials for you.  You will only incur charges as you create topics and produce and consume data.

or something like that...

ccloud/docs/ccloud-stack.rst
```diff
@@ -30,8 +30,33 @@ Caution
 =======
 
 This utility uses real |ccloud| resources.
+It is intended to be a quick way to create resources in |ccloud| with correct credentials and permissions, useful as a starting point from which you can then use for learning, extending, and building other examples.
+If you just run ``ccloud-stack`` without enabling |ccloud| ksqlDB, then there is no billing charge (but as soon as you produce data to the |ak| cluster or provision any other fully-managed services, there may be charges).
+If you run ``ccloud-stack`` with enabling |ccloud| ksqlDB, then you will begin in accrue charges.
```
rspurgeon:
```suggestion
If you run ``ccloud-stack`` with enabling |ccloud| ksqlDB, then you will begin to accrue charges.
```

rspurgeon (approved):
LGTM, thanks

bechols:
Yep, shouldn't be any costs without ksqlDB since this is creating a Basic cluster which doesn't have an uptime charge and not actually using the cluster at all.

caleb-grillo (approved):

ccloud/docs/ccloud-stack.rst
```diff
@@ -30,8 +30,33 @@ Caution
 =======
 
 This utility uses real |ccloud| resources.
+It is intended to be a quick way to create resources in |ccloud| with correct credentials and permissions, useful as a starting point from which you can then use for learning, extending, and building other examples.
+If you just run ``ccloud-stack`` without enabling |ccloud| ksqlDB, then there is no billing charge until you produce data to the |ak| cluster or provision any other fully-managed services.
+If you run ``ccloud-stack`` with enabling |ccloud| ksqlDB, then you will begin to accrue charges immediately.
```
matt-mangia, line 35: These 3 lines are great!

matt-mangia (approved):
I think this is super helpful & clarifies a lot for me in using the tool! Thank you :)

ccloud/docs/ccloud-stack.rst
ybyzek: Excellent, thanks for reviewing @matt-mangia

rspurgeon (approved):
LGTM

PR #839, opened 2020-10-15
ST-3461: Update to work with nano versioning

elismaga:
### Description 

This PR updates this repo so that it will work after we roll out nano versioning. This will be merged as part of the nano versioning roll out.

I updated the reference to the parent version to use a version range so that it gets the most recent 6.1.0 artifacts. I also updated all the CP dependencies to use the new version properties.

This repo will still build as version 6.1.0-SNAPSHOT though. If we wish to start building nano versions of these artifacts we can make that update in the future, but it is not needed now.

### Author Validation

I am validating the changes by making sure we get a successful build of this PR.

### Reviewer Tasks


awalther28:
@rspurgeon might disagree but @elismaga I think this PR should be against `6.0.0-post`. We use `pint merge` for versioning on this repo. Also I did run `mvn clean package` in `mulit-datacenter/` and `connect-streams-pipeline/`, both failed to compile. Maybe I'm missing something? I still need to watch the nano version info session. @rspurgeon is the 'veteran' team member I hinted to on slack who is out today, I'll sync up with him tomorrow morning.

elismaga:
@awalther28 Nano versioning was only introduced in 6.1.0, so it wouldn't work on any older versions such as 6.0.0, so we shouldn't merge this to 6.0.0-post. 

I will try building those projects and figure out what the issue is.

elismaga:
@awalther28 I was able to build both of those maven projects. Could you slack me the errors you are getting

awalther28:
@elismaga 🤦 I didn't have confluent’s internal artifact repository setup for maven, hence why I couldn't get those demos to build yesterday. **I currently have them building just fine**. Also today I learned what all the tools team does. Truly my apologies, still learning the ropes here at Confluent 😅

elismaga:
> @elismaga 🤦 I didn't have confluent’s internal artifact repository setup for maven, hence why I couldn't get those demos to build yesterday. **I currently have them building just fine**. Also today I learned what all the tools team does. Truly my apologies, still learning the ropes here at Confluent 😅

@awalther28 No worries, and welcome aboard.

ybyzek:
@elismaga thank you for proactively submitting this PR.  

Just to force the linkage to a similar PR, would like to note that equivalent changes where made in `kafka-streams-examples`: https://github.com/confluentinc/kafka-streams-examples/pull/356 (including the reference to `blueway`).

Questions:

1. @elismaga it looks like https://github.com/confluentinc/kafka-streams-examples/pull/356 sets at the top (just under `<packaging>jar</packaging>`) `<version>6.1.0-0</version>` whereas this PR sets it to `<version>6.1.0-SNAPSHOT</version>`.  Why are these different, should they be the same, and if so, which one is right?

2. As @rspurgeon noted, there are also some `build.gradle` files.  Do we know what lines/values are impacted in those files, what values they should be set to, and whether automation will automatically bump them?

elismaga:
@ybyzek 

1. The kafka-streams-examples repo is part of the CP pipeline and release process which now uses nano versioning so that is why the version was changed to 6.1.0-0. Since this repo is not part of CP it does not use nano versioning yet, but in the future we can try and get this repo to produce nano versions if you want. Since it is not configured to use nano versioning we still need to produce snapshot versions. That is why the version of the examples repo was not changed. To recap, it is correct that examples still builds snapshots, and kafka-streams-examples builds nano versions such as 6.1.0-0.

2. Besides from the pom files all the other references to versions I found were mostly in .env files and from what I could tell they shouldn't be changed because it looked to me like they were referencing the version of the artifacts being built in this repo which are still 6.1.0-SNAPSHOT. When I searched for gradle files I didn't see any that contained version information. Maybe I missed something though, so can you point me to a specific gradle file that you are think of?

3. If you are using the release tools scripts such as tbr, or the release engineering team has been doing the branch cuts for new releases for you then yes, they will bump the versions for you just like they have been doing. They have been adjusted for nano versioning so they will update the version ranges that were added here.

elismaga:
@ybyzek I was just thinking about this a little more and discussing with @andrewegel and since this repo does not have a jenkins build there really isn't anything that needs to be done in order to convert this repo to produce nano versioned artifacts except to change the version it produces to 6.1.0-0, and change all the references in the config files to use that version. This will actually make the tbr scripts handle the version bumps better as well. I am going to add those changes to this PR now as well.

rspurgeon:
@elismaga re:
> When I searched for gradle files I didn't see any that contained version information. Maybe I missed something though, so can you point me to a specific gradle file that you are think of?

./cloud/java-springboot/build.gradle
./cloud/groovy/build.gradle

Please note: ` implementation 'io.confluent:kafka-streams-avro-serde:5.4.0'`

rspurgeon (approved):
I've spot checked some of the examples this touches and things seem to function as expected.  I'm still a little unclear what will happen at branch cut / release time with these values and related tooling automation, however, assuming that all works as expected this LGTM

andrewegel (approved):
LGTM, we can work through version bump automation issues through the CP 6.1 cycle this quarter - I suspect we'll see bump issues after the branch-cut/after a nano-versioning release, so I'll keep an eye out for any ST Jiras relating to this process change.. 

~~One question I had @elismaga is what these values will be when it comes time to creating a CP 6.1 On-Prem release (essentially what these target), do we pin the nanoversion, or do we keep the range component in there?~~

Spoke to Eli offline about this - Turns out `6.1.0` will be greater than `6.1.0-N`, which the former `6.1.0` will likely be whats set here when we cut CP Releases.

elismaga:
@rspurgeon The branch cut / release tooling has been updated and should work with this, but if any issues arise I will work with Andrew to fix them.

PR #841, opened 2020-10-22
DEVX-2194: double quotes exported vars in ccloud generate script

rspurgeon:
### Description 
Prior to this change the quoting of exported variables from this script would be improper resulting in configuration errors, specificially observed in docker usage.

### Author Validation

- [X] microservices-orders

### Reviewer Tasks
Validate any examples that utilize ccloud-stack


ybyzek (approved):
@rspurgeon LGTM, I validated microservices-orders to ccloud.  Thank you for fixing the collateral damage from https://github.com/confluentinc/examples/pull/836

PR #849, opened 2020-10-29
 DEVX-2219: Add validate.sh for quick validation of example: multi-datacenter

ybyzek:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2219

### Author Validation

- [x] multi-datacenter

### Reviewer Tasks

No need to validate, just review the concept.  This is the first of a few that we will implement for all demos.  Note: the validation is meant to be simple with low investment.  cc: @sdanduConf WDYT?


rspurgeon (approved):
LGTM

PR #862, opened 2020-11-17
Update README.md

rmoff:
fix typo


ybyzek (approved):
Good catch, thanks @rmoff

PR #876, opened 2020-12-11
DEVX-2334: fix microservices demo and JAR location

ybyzek:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2334

### Author Validation

- [x] microservices-orders: Docker and local install

### Reviewer Tasks

- [x] microservices-orders: Docker and local install

### DIFF
--- a/microservices-orders/scripts/run-services.sh
+++ b/microservices-orders/scripts/run-services.sh
@@ -5,8 +5,7 @@ set -m
 BOOTSTRAP_SERVERS=${BOOTSTRAP_SERVERS:-broker:9092}
 SCHEMA_REGISTRY_URL=${SCHEMA_REGISTRY_URL:-http://schema-registry:8081}
 RESTPORT=${RESTPORT:-18894}
-JAR_DEFAULT="/usr/share/java/kafka-streams-examples/"$(ls /usr/share/java/kafka-streams-examples/ | grep standalone | awk '{print $NF;}')
-JAR=${JAR:-$JAR_DEFAULT}
+JAR=${JAR:-"/usr/share/java/kafka-streams-examples/kafka-streams-examples-*-standalone.jar"}
 PIDS=()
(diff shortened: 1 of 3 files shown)


microservices-orders/scripts/run-services.sh
```diff
@@ -5,7 +5,8 @@ set -m
 BOOTSTRAP_SERVERS=${BOOTSTRAP_SERVERS:-broker:9092}
 SCHEMA_REGISTRY_URL=${SCHEMA_REGISTRY_URL:-http://schema-registry:8081}
 RESTPORT=${RESTPORT:-18894}
-JAR=${JAR:-"/usr/share/java/kafka-streams-examples/kafka-streams-examples-$CONFLUENT-standalone.jar"}
+JAR_DEFAULT="/usr/share/java/kafka-streams-examples/"$(ls /usr/share/java/kafka-streams-examples/ | grep standalone | awk '{print $NF;}')
```
ybyzek: @rspurgeon any suggestions on making this prettier?

rspurgeon: I don't think we need to be specific about the actual jar file name, we can use wildcards:

JAR=${JAR:-"/usr/share/java/kafka-streams-examples/kafka-streams-examples-*.jar"}

But for the local build part, I think we should add a `mvn clean` stage in the `get_and_compile_kafka_streams_examples` function so that it will clean out any old build artifacts before running.  Otherwise, the wildcard might be an issue with previous artifacts.

ybyzek: @rspurgeon great suggestion.  Added in latest commit https://github.com/confluentinc/examples/pull/876/commits/14b6749de59bb9a32754be1b747c4f60c4ff7995

rspurgeon (approved):
LGTM, validated CP local and Docker

PR #877, opened 2020-12-11
DEVX-2324: Two source datagen connector went in degraded state (task …

ybyzek:
…failed) state just after deployment

### Description 

https://confluentinc.atlassian.net/browse/DEVX-2324

History: the datagen connectors used to run indefinitely because the iterations parameter was misspelled. The spelling error was fixed in https://github.com/confluentinc/examples/pull/871/files#diff-1f0220c18c3cb85a6616f5532f4f4ef55314d53f5a680faaf990dc48965661c1R133 , and consequently the datagen connector stops after iterations number is reached and generates those log messages – this is expected and working as designed. However, the docs were not updated to reflect this.

Resolution: remove `iterations` parameter completely to avoid connector log messages, connectors won't show degraded state (which might be confusing for users), docs won't need to change

### Author Validation

- [x] clickstream

### Reviewer Tasks

- [x] clickstream


bbejeck (approved):
LGTM!

Validated by running clickstream via `start.sh` and manual steps

PR #902, opened 2021-01-11
DEVX-2256: fix: removes unnecessary key converter configurations

rspurgeon:
These configurations are deprecated, and the current values in all cases
matches the values of the default in the new names.  Removing these should
have no effect on behavior.

### Description 
Removes deprecated and unecessary configurations

There should be no change in behavior with these removals.

### Author Validation

I validated the clients/cloud/kafka-connect-datagen example which uses connect to connect to CCloud

### Reviewer Tasks
Reivewer could test an alternate demo which utilizes connect that had the configuraiton removed.


awalther28 (approved):
I validated the changes by running the multi-datacenter demo. Behavior looks right to me. I approve.

ybyzek (approved):
@rspurgeon LGTM (code inspection only...confidence is high given @awalther28 's validation and the related PR to cp-demo)

PR #913, opened 2021-01-17
Update config.env in accordance with cp build tags to fix issues like DEVEX-2435

mohitgrewal:
### Description 
Update config.env in accordance with cp build tags to fix issues like DEVEX-2435

https://confluentinc.atlassian.net/browse/DEVX-2435
CP builds with RC build are moved to have 6.1.0 tag for jar files. CONFLUENT variable in examples/utils/config.env still have 6.1.0-0 values which does not match cp jar tags. As we approach 6.1 release final tags will be 6.1.0. This PR updates these variables in config.env in accordance with cp tags.

### Author Validation

- [ ] multi-datacenter

### Reviewer Tasks

### DIFF
--- a/utils/config.env
+++ b/utils/config.env
@@ -5,8 +5,8 @@
 # over time to construct other
 # values required by this repository
 #####################################################
-CONFLUENT=6.1.0-0
-CONFLUENT_DOCKER_TAG=6.1.x-latest
+CONFLUENT=6.1.0
+CONFLUENT_DOCKER_TAG=6.1.0-latest
 CONFLUENT_SHORT=6.1
 CONFLUENT_PREVIOUS=""
 CONFLUENT_RELEASE_TAG_OR_BRANCH=6.1.x


ybyzek:
@andrewegel for clarification, shouldn't tooling be doing this file edit automatically when 6.1.0 is GA?  If yes, then is it the case that this file should be only _locally_ edited when testing RC images for now?

Update: @mohitgrewal https://github.com/confluentinc/examples/pull/914 may be a better long-term fix than this PR (which circumvents the tooling automation).  I propose we close this in favor of #914

andrewegel:
> @andrewegel for clarification, shouldn't tooling be doing this file edit automatically when 6.1.0 is GA? If yes, then is it the case that this file should be only locally edited when testing RC images for now?

Only after the release is done will this be set to a concrete version (6.1.0) in the `-post` branch (6.1.0-post) - The time between when an RC is cut and release, yes you have to manually edit the tag to `6.1.0-rc210114220812-latest` to test the RC tag (when a release occurs, the `6.1.0-rc210114220812-latest` will be re-tagegd as `6.1.0`)

PR #916, opened 2021-01-18
DEVX-2443: Tag Confluent Cloud examples by stuffing identifiers into …

ybyzek:
…names or descriptions of CCloud resources

### Description 

https://confluentinc.atlassian.net/browse/DEVX-2443

1. Add example name to env name
2. Add user email to SA description

### Author Validation

### Reviewer Tasks

### DIFF
--- a/ccloud/beginner-cloud/start.sh
+++ b/ccloud/beginner-cloud/start.sh
@@ -24,7 +24,7 @@ check_jq || exit 1
 # Create a new environment and specify it as the default
 ##################################################
 
-ENVIRONMENT_NAME="demo-script-env"
+ENVIRONMENT_NAME="demo-env-000000-beginner-cli"
 echo -e "\n# Create a new Confluent Cloud environment $ENVIRONMENT_NAME"
 echo "ccloud environment create $ENVIRONMENT_NAME -o json"
 OUTPUT=$(ccloud environment create $ENVIRONMENT_NAME -o json)
--- a/ccloud/beginner-cloud/start.sh
+++ b/ccloud/beginner-cloud/start.sh
@@ -24,7 +24,7 @@ check_jq || exit 1
 # Create a new environment and specify it as the default
 ##################################################
 
-ENVIRONMENT_NAME="demo-env-000000-beginner-cli"
+ENVIRONMENT_NAME="ccloud-stack-000000-beginner-cli"
 echo -e "\n# Create a new Confluent Cloud environment $ENVIRONMENT_NAME"
 echo "ccloud environment create $ENVIRONMENT_NAME -o json"
 OUTPUT=$(ccloud environment create $ENVIRONMENT_NAME -o json)
--- a/utils/ccloud_library.sh
+++ b/utils/ccloud_library.sh
@@ -293,6 +293,17 @@ function ccloud::validate_schema_registry_up() {
   return 0
 }
 
+function ccloud::get_environment_id_from_service_id() {
+  SERVICE_ACCOUNT_ID=$1
+
+  ENVIRONMENT_NAME_PREFIX=${ENVIRONMENT_NAME_PREFIX:-"demo-env-$SERVICE_ACCOUNT_ID"}
+  ENVIRONMENT=$(ccloud environment list -o json | jq -r 'map(select(.name | startswith("'"$ENVIRONMENT_NAME_PREFIX"'"))) | .[].id')
+
+  echo $ENVIRONMENT
+
+  return 0
+}
+
 
 function ccloud::create_and_use_environment() {
   ENVIRONMENT_NAME=$1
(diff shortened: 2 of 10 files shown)


utils/ccloud_library.sh
```diff
@@ -356,7 +356,8 @@ function ccloud::maybe_create_and_use_cluster() {
 function ccloud::create_service_account() {
   SERVICE_NAME=$1
 
-  OUTPUT=$(ccloud service-account create $SERVICE_NAME --description $SERVICE_NAME  -o json)
+  CCLOUD_EMAIL=$(ccloud prompt -f '%u')
+  OUTPUT=$(ccloud service-account create $SERVICE_NAME --description "$EXAMPLE run by $CCLOUD_EMAIL"  -o json)
```
javabrett: The other thing we are typically missing are the resource creation dates.  Would it be possible to add a datetime stamp, either here in the service-account description, or in the environment name (which to be fair would be getting quite busy then).

ybyzek: @javabrett  I can get env creation date...so if that data is useful for you, I can share the link (DM in Slack please).  But stepping back, we should think about how far we want to go in making up for things the product itself doesn't do yet. 😄

javabrett (approved):
LGTM and I like the outcome of the change.  I ran `cp-quickstart` and my environment is named `demo-env-<NNNNNN>-cp-quickstart`, and my service account has:

```
  <NNNNNN> | demo-app-28283    | SA for cp-quickstart run by
         |                   | my.email@confluent.io
```

This works well and improves demo identification.

ccloud/beginner-cloud/start.sh
```diff
@@ -24,7 +24,7 @@ check_jq || exit 1
 # Create a new environment and specify it as the default
 ##################################################
 
-ENVIRONMENT_NAME="demo-script-env"
+ENVIRONMENT_NAME="demo-env-000000-beginner-cli"
```
rspurgeon: What does `000000` represent here?

An additional question, considering we are refactoring this bit, could it make sense to remove the `env` from `demo-env`?  Based on consideration that an environment already categorizes itself as an environment. So do we need the word "env" when naming environments, given that's what they are the name may be redundant. 

Potentially there could be a more useful category instead of `env`, like `ia` or `stack`

utils/ccloud_library.sh
```diff
@@ -293,6 +293,17 @@ function ccloud::validate_schema_registry_up() {
   return 0
 }
 
+function ccloud::get_environment_id_from_service_id() {
+  SERVICE_ACCOUNT_ID=$1
+
+  ENVIRONMENT_NAME_PREFIX=${ENVIRONMENT_NAME_PREFIX:-"demo-env-$SERVICE_ACCOUNT_ID"}
+  ENVIRONMENT=$(ccloud environment list -o json | jq -r 'map(select(.name | startswith("'"$ENVIRONMENT_NAME_PREFIX"'"))) | .[].id')
```
rspurgeon: I don't think the intention of this function is to modify a global instance of `ENVIRONMENT`, but this will. I suggest we either echo the result of the previous command without assignment, or we use the `local` modifier (and maybe a different var name) for the variable to avoid mutating the global version of this.

ccloud/beginner-cloud/start.sh
ybyzek: 1. `xxxxxx` is the service account ID.  Since this tutorial is a bit different (env created before SA), it uses `000000`.
2. Agree, will remove `env`.  I don't think there's use in adding `ia` or `stack` though...unless there is concern about user-defined environments that start with `demo-`. WDYT?

utils/ccloud_library.sh
```diff
@@ -356,7 +367,8 @@ function ccloud::maybe_create_and_use_cluster() {
 function ccloud::create_service_account() {
   SERVICE_NAME=$1
 
-  OUTPUT=$(ccloud service-account create $SERVICE_NAME --description $SERVICE_NAME  -o json)
+  CCLOUD_EMAIL=$(ccloud prompt -f '%u')
+  OUTPUT=$(ccloud service-account create $SERVICE_NAME --description "SA for $EXAMPLE run by $CCLOUD_EMAIL"  -o json)
```
rspurgeon, line 371: I'm curious if we might benefit from using the description fields to put in more parsable text? Also do we need the word "SA" when it's already attached to a Service Account?

FYC instead of:

"SA for example run by user"

"user:example" would be easier to parse if needed by a machine and I'm not sure that the other wording does all that much for the human reader as they will see the description in the context of listing of service accounts.

Just my opinion...

ccloud/beginner-cloud/start.sh
rspurgeon: I'm good w/ just `demo` and no additional category. 👍

ybyzek: @rspurgeon I decided to make it `ccloud-stack-` prefix.  Reasons:

1. Side-step opinion whether its purpose is demo/example/production/other
2. More explicit about how it was created

utils/ccloud_library.sh
ybyzek: @rspurgeon `SA for ` language was intentionally borrowed from the ksqlDB description when a new app is provisioned.  We don't have to be consistent with that, but that was the starting point.  @rspurgeon @javabrett preferences for consistency ksqlDB over parsability?

utils/ccloud_library.sh
ybyzek: @rspurgeon thoughts on latest change?

rspurgeon (approved):
LGTM

PR #918, opened 2021-01-20
DEVX-2434: examples/microservices 6.1: ksql query set up is failing for docker example

awalther28:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2434

This ticket points out issues with a static sleep and suggests we move to a poll prior to running the ksqlDB statements in the ksql-query-setup container. While ksqlDB has [bash for waiting for the ksqlDB server to be ready](https://docs.ksqldb.io/en/latest/operate-and-deploy/installation/install-ksqldb-with-docker/#wait-for-an-http-endpoint-to-be-available), I found that this snippet of bash will exit the loop prematurely. It will exit prior to the server being in a "RUNNING" state and prior to the necessary topics being made.

I suggest we wait until the ksqlDB server says it's running and the the topics the statements will use are present, which is present in the PR. 

### Author Validation

 - [x] microservices-orders --> docker only

### Reviewer Tasks

 - [ ] microservices-orders --> docker only

### DIFF
--- a/microservices-orders/run-ksql.sh
+++ b/microservices-orders/run-ksql.sh
@@ -1,7 +1,12 @@
 #!/bin/bash
 
-echo "Sleeping 300 seconds until ksqlDB server starts and topics have data"
-sleep 300
+echo -e "\nWaiting for ksqlDB and topics to be available prior to running ksqlDB statements\n"
+while [[ $(curl -s http://ksqldb-server:8088/info) != *RUNNING* ||  $(curl -s http://schema-registry:8081/subjects) != *platinum* ]]
+do 
+    sleep 5
+    echo "Sleeping for 5 seconds, waiting for ksqlDB and topics to be available"
+done
+
 
 # Create KSQL queries
 ksql http://ksqldb-server:8088 <<EOF
--- a/microservices-orders/run-ksql.sh
+++ b/microservices-orders/run-ksql.sh
@@ -1,16 +1,15 @@
 #!/bin/bash
+source /tmp/helper.sh
 
-echo -e "\nWaiting for ksqlDB and topics to be available prior to running ksqlDB statements\n"
-while [[ $(curl -s http://ksqldb-server:8088/info) != *RUNNING* \
-             ||  $(curl -s http://schema-registry:8081/subjects) != *customers* \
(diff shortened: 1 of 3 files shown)


microservices-orders/run-ksql.sh
```diff
@@ -1,7 +1,14 @@
...
-sleep 300
+echo -e "\nWaiting for ksqlDB and topics to be available prior to running ksqlDB statements\n"
+while [[ $(curl -s http://ksqldb-server:8088/info) != *RUNNING* \
+             ||  $(curl -s http://schema-registry:8081/subjects) != *customers* \
+             ||  $(curl -s http://schema-registry:8081/subjects) != *orders* ]]
+do 
+    sleep 5
+    echo "Sleeping for 5 seconds, waiting for ksqlDB and topics to be available"
+done
+
```
rspurgeon: Nice change @awalther28 .  What do you think about leveraging the existing, and available `retry` function?  It has 'abort' capability.  Maybe you could wrap up the checks into a function inside this script and then invoke it with `retry` or figure out a slick way to pass in these calls to the `retry` directly.  There is an example of `retry` usage in the `read-topics-docker.sh` script and it's defined in `utils/helper.sh`

rspurgeon: Or maybe that's not available inside the docker container.... so that might not work.  If not, then I'm wondering if we should add a max wait equivilant here?

rspurgeon: Another consideration for you, if you decided to break this into a function, you could remove the redundant curl calls to SR and stuff the result into a variable, then check the value of the variable instead of invoke `curl` twice for the same info.  There also might be a solution using a REGEX but that could be more work than it's worth.

awalther28: @rspurgeon I have applied some of the changes you suggested: namely creating functions, limiting curl usage, and leveraging `utils/helper.sh retry`. Bash is not a strong suit of mine, definitely open to any suggestions you have to make things better.

awalther28: @rspurgeon bump 🙂

rspurgeon: nice :-)

rspurgeon (approved):
LGTM

PR #920, opened 2021-01-22
DEVX-2463: ccloud-stack: generate configs failing errors

awalther28:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2463

Without this PR, users using `zsh` will hit this failure (`ccloud::generate_configs:39: no matches found: -F["']`) when generating configs with ccloud. @ybyzek said we only need to account for CCloud credentials produced by CCloud CLI being wrapped in single quotes, thus we don't need to look for double quotes. Removing that resolves the error for `zsh` users. 

`bash` users don't encounter any errors. 

### Author Validation

 - [x] ccloud --> generating configs with`ccloud/cloud-generate-cp-configs.sh` and `ccloud::generate_configs` in `bash` and `zsh`

### Reviewer Tasks

 - [ ] ccloud --> generating configs with`ccloud/cloud-generate-cp-configs.sh` and `ccloud::generate_configs` in `bash` and `zsh`

### DIFF
--- a/utils/ccloud_library.sh
+++ b/utils/ccloud_library.sh
@@ -1146,8 +1146,8 @@ function ccloud::generate_configs() {
   SASL_JAAS_CONFIG=$( grep "^sasl.jaas.config" $CONFIG_FILE | cut -d'=' -f2- )
   SASL_JAAS_CONFIG_PROPERTY_FORMAT=${SASL_JAAS_CONFIG/username\\=/username=}
   SASL_JAAS_CONFIG_PROPERTY_FORMAT=${SASL_JAAS_CONFIG_PROPERTY_FORMAT/password\\=/password=}
-  CLOUD_KEY=$( echo $SASL_JAAS_CONFIG | awk '{print $3}' | awk -F[\"\'] '$0=$2' )
-  CLOUD_SECRET=$( echo $SASL_JAAS_CONFIG | awk '{print $4}' | awk -F[\"\'] '$0=$2' )
+  CLOUD_KEY=$( echo $SASL_JAAS_CONFIG | awk '{print $3}' | awk -F"'" '$0=$2' )
+  CLOUD_SECRET=$( echo $SASL_JAAS_CONFIG | awk '{print $4}' | awk -F"'" '$0=$2' )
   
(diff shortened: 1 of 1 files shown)


utils/ccloud_library.sh
```diff
@@ -1240,6 +1240,7 @@ function ccloud::generate_configs() {
   ################################################################################
   KSQLDB_SERVER_DELTA=$DEST/ksqldb-server-ccloud.delta
   echo "$KSQLDB_SERVER_DELTA"
+  rm -f $KSQLDB_SERVER_DELTA
```
rspurgeon, line 1243: IMO we do not need `-f` here.  I tested this command without, and considering the user previously created the file in a folder that they have proper permissions on, the `-f` isn't required.  So introducing the force option seems like an unecessary risk, in case something important (oddly) happen to get into the KSQLDB_SERVER_DELTA variable.

If we want to silence error messages if the file doesn't exist, i suggeste we use `rm $KSQLDB_SERVER_DELTA 2>/dev/null`

awalther28: I think that's a fair point @rspurgeon . Can I assume you'd want to alter how the other delta files are deleted in this function? 
```
rm -f $REPLICATOR_PRODUCER_DELTA
rm -f $SR_CONFIG_DELTA
rm -f $INTERCEPTORS_CONFIG_FILE
rm -f $KSQL_DATAGEN_DELTA
rm -f $C3_DELTA
rm -f $METRICS_REPORTER_DELTA
rm -f $REST_PROXY_DELTA
rm -f $CONNECT_DELTA
rm -f $CONNECTOR_DELTA
... (list goes on) ...
```
Personally I think if we change this one then we should change all of them.

rspurgeon: Thanks @awalther28 ... If you're up for changing all in this function, I agree.  IMO it's ok to not introduce any new usage of `-f` to start, but I don't think we should be force removing files on our users machines generally.  But starting here seems great to me

rspurgeon (approved):
LGTM with the consideration to remove the -f from the new `rm` addition at your discretion

PR #921, opened 2021-01-22
DEVX-2463-cont: replace -f removal of files

awalther28:
### Description 

In https://github.com/confluentinc/examples/pull/920 @rspurgeon suggested 
>IMO we do not need -f here. I tested this command without, and considering the user previously created the file in a folder that they have proper permissions on, the -f isn't required. So introducing the force option seems like an unecessary risk, in case something important (oddly) happen to get into the KSQLDB_SERVER_DELTA variable.

>If we want to silence error messages if the file doesn't exist, i suggeste we use rm $KSQLDB_SERVER_DELTA 2>/dev/null

To be cautious we should refrain from force removing files in our demos when we don't have to. This PR changes for `rm -f` to `rm $KSQLDB_SERVER_DELTA 2>/dev/null`. I validated this by running ccloud-stack-create (generating configs with both `bash` and `zsh`) and running the microservices demo. 

### Author Validation

- [ ] ccloud 

### Reviewer Tasks

- [ ] ccloud 
- [ ] other cloud demos that use ccloud-stack-create


rspurgeon (approved):
LGTM, thanks!

PR #924, opened 2021-01-28
update sr configs to use basic.auth.user.info instead of schema.regis…

laura-brouckman:
…try.basic.auth.user.info

From this ticket: https://confluentinc.atlassian.net/browse/CPG-606

### Description 

### Author Validation

### Reviewer Tasks

### DIFF
--- a/clients/docs/includes/configs/cloud/java-sr.config
+++ b/clients/docs/includes/configs/cloud/java-sr.config
@@ -12,4 +12,4 @@ acks=all
 # Required connection configs for Confluent Cloud Schema Registry
 schema.registry.url=https://{{ SR_ENDPOINT }}
 basic.auth.credentials.source=USER_INFO
-schema.registry.basic.auth.user.info={{ SR_API_KEY }}:{{ SR_API_SECRET }}
+basic.auth.user.info={{ SR_API_KEY }}:{{ SR_API_SECRET }}
--- a/clients/docs/includes/configs/cloud/librdkafka-sr.config
+++ b/clients/docs/includes/configs/cloud/librdkafka-sr.config
@@ -8,4 +8,4 @@ sasl.password={{ CLUSTER_API_SECRET }}
 # Confluent Cloud Schema Registry
 schema.registry.url=https://{{ SR_ENDPOINT }}
 basic.auth.credentials.source=USER_INFO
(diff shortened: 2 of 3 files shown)


ybyzek:
@laura-brouckman thanks for the PR

Have these changes been validated: running REST proxy with this new config?  They differ from https://github.com/confluentinc/cp-all-in-one/blob/6.0.1-post/cp-all-in-one-cloud/docker-compose.yml#L158, which currently work but may also need to change?

ybyzek:
Update: I tested the docker-compose with the similar change, and it also works.  I'm confused now why the internally linked jiras suggested that the current configs did not work.

laura-brouckman:
I didn't actually test REST Proxy with these changes- in the Cloud UI, this change would only affect the client configs (so not configs for other CP components like rest proxy)

 Recently I did change the rest proxy config (in the cloud UI) based a separate ticket so now with SR enabled, it includes the lines:
Screen Shot 2021-01-28 at 10 31 40 AM
Before it was missing the `client.` before those settings

But if these files are being used in our docs/examples repo for things other than just configuring clients i can try to go test more things

ybyzek:
@laura-brouckman sorry I was confused: I thought this PR was addressing CPG-782 which I just filed 2 days ago, but I now see it is addressing CPG-606.  This change is ok then.  Please note: there are other places in this repo where the equivalent change needs to be made, and that is being tracked in https://confluentinc.atlassian.net/browse/DEVX-2255

ybyzek (approved):
LGTM thanks @laura-brouckman

laura-brouckman:
thanks for reviewing @ybyzek! I can't actually merge (no permissions) so if it looks alright, could you hit the button? 🙏

ybyzek:
@laura-brouckman my apologies, I missed one more important point -- the target branch.  If we want this to be live in CCloud UI, then we need to retarget this PR to merge to `6.0.1-post` (not `master`).  Afterwards, I'll update `latest` to point to that commit in `6.0.1-post`, which is what CCloud pulls from IIRC.

ybyzek:
```
* 3592eff8 (HEAD -> latest, origin/latest, origin/HEAD, origin/6.0.1-post, 6.0.1-post) update sr configs to use basic.auth.user.info instead of schema.registry.basic.auth.user.info (#924)
```

All set!  Thank you again @laura-brouckman

PR #935, opened 2021-02-12
DEVX-2166: use KIP-464 for default replication factor from brokers: c…

ybyzek:
… client

### Description 

https://confluentinc.atlassian.net/browse/DEVX-2166

Use default RF value from broker, to allow the code to be used to clusters where RF=3 or RF=1

### Author Validation

- [x] clients/cloud: validated c example

### Reviewer Tasks

- [ ] clients/cloud: validated c example

### DIFF
--- a/clients/cloud/c/common.c
+++ b/clients/cloud/c/common.c
@@ -124,12 +124,13 @@ int create_topic (rd_kafka_t *rk, const char *topic,
         rd_kafka_event_t *rkev;
         const rd_kafka_CreateTopics_result_t *res;
         const rd_kafka_topic_result_t **restopics;
+        const int replication_factor_or_use_default = -1;
         size_t restopic_cnt;
         int ret = 0;
 
         fprintf(stderr, "Creating topic %s\n", topic);
 
-        newt = rd_kafka_NewTopic_new(topic, num_partitions, -1,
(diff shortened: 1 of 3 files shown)


clients/cloud/c/common.c
```diff
@@ -129,7 +129,7 @@ int create_topic (rd_kafka_t *rk, const char *topic,
 
         fprintf(stderr, "Creating topic %s\n", topic);
 
-        newt = rd_kafka_NewTopic_new(topic, num_partitions, replication_factor,
+        newt = rd_kafka_NewTopic_new(topic, num_partitions, -1,
```
edenhill: `.., num_partitions, -1/*use broker default RF*/,`

edenhill (approved):
The README should be updated that this code now requires librdkafka v1.6.0 or later.

clients/cloud/c/common.c
rspurgeon: Another idea could be to use a const

```
/* Signals to use the broker default */
const int DEFAULT_REPLICATION_FACTOR = -1;
```

and then
```
        newt = rd_kafka_NewTopic_new(topic, num_partitions, DEFAULT_REPLICATION_FACTOR,
                                     errstr, sizeof(errstr));
```

rspurgeon (approved):
Left a minor comment, otherwise LGTM tested locally against CC following the doc instructions and worked as advertised.

PR #938, opened 2021-02-16
DEVX-2513: update catch for validate_logged_in_ccloud_cli

ybyzek:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2513

Note: I didn't catch on return code because I was thinking that the test command could fail for other reasons and error out, not just because not logged in.  

### Author Validation

- [x] cp-quickstart

### Reviewer Tasks

- [ ] cp-quickstart


rspurgeon (approved):
LGTM

PR #940, opened 2021-02-17
DEVX-2518: update caution note for running CCloud examples

ybyzek:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2518

### Author Validation

### Reviewer Tasks

### DIFF
--- a/utils/ccloud_library.sh
+++ b/utils/ccloud_library.sh
@@ -34,16 +34,17 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 
 function ccloud::prompt_continue_ccloud_demo() {
   echo
-  echo "--------------------------------------------------------------------------------------------"
+  echo "-------------------------------------------------------------------------------------------"
   echo "Any Confluent Cloud example uses real Confluent Cloud resources that may be billable,"
   echo "including connectors and ksqlDB applications that may have hourly charges."
   echo
+  echo "At the end of this script, it will show a command to destroy all Confluent Cloud resources."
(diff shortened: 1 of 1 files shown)


utils/ccloud_library.sh
```diff
@@ -34,16 +34,16 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
...
   echo
-  echo "To avoid unexpected charges, carefully evaluate the cost of resources before you start."
-  echo "After you are done running a Confluent Cloud example, destroy all Confluent Cloud resources"
-  echo "to avoid accruing hourly charges for services and verify that they have been deleted."
-  echo "---------------------------------------------------------------------------------------------"
+  echo "Use Confluent Cloud promo code C50INTEG to receive \$50 free usage, which should"
+  echo "sufficiently cover one day of running this example, beyond which you may be billed"
+  echo "for the services until you destroy the resources."
+  echo
+  echo "Use the provided script to destroy all resources. Always verify that they have been deleted."
```
rspurgeon: FYC.  I'm just trying to inform the user where the the destroy command will be presented to them later so they aren't looking for it here while reading this.

```suggestion
  echo "After startup, this demo will provide a command to destroy all the demo resources. After running the command, use the Confluent Cloud UI or the `ccloud` CLI to verify the demo resources have been deleted."
```

ybyzek: Good idea.  I hadn't done that at first because I wasn't sure every example did that, but now I checked and see that they do!  So, I will rework the text to capture this suggestion (but will fix line breaks).

rspurgeon (approved):
LGTM

PR #946, opened 2021-02-23
DEVX-2461: ccloud-stack: ksqlDB app creation reuses SA

rspurgeon:
### Description 

ccloud-stack: ksqlDB app creation reuses SA. There are other fixes as well related to docs, some minor functional reordering and validtions on functions.

### Author Validation

- [X] Documentation
- [X] ccloud/ccloud-stack (script and source function)
- [X] cp-quickstart cloud

### Reviewer Tasks

- [ ] Any demo which uses ccloud-stack

### DIFF
--- a/ccloud/docs/ccloud-stack.rst
+++ b/ccloud/docs/ccloud-stack.rst
@@ -142,7 +142,6 @@ Create a ccloud-stack
 
    .. code-block:: text
 
-ACLs in this cluster:
         ServiceAccountId | Permission |    Operation     |     Resource     |             Name             |   Type
       +------------------+------------+------------------+------------------+------------------------------+----------+
         User:186588      | ALLOW      | DESCRIBE_CONFIGS | GROUP            | *                            | LITERAL
--- a/ccloud/docs/ccloud-stack.rst
+++ b/ccloud/docs/ccloud-stack.rst
@@ -54,7 +54,7 @@ By default, the |ccloud| ksqlDB app is not created with ``ccloud-stack``, you ha
    ccloud api-key create --service-account $SERVICE_ACCOUNT_ID --resource $RESOURCE -o json    // for schema-registry
 
    # By default, ccloud-stack does not enable Confluent Cloud ksqlDB, but if you explicitly enable it:
-   ccloud ksql app create --cluster $CLUSTER --api-key "$kafka_api_key" --api-secret "$kafka_api_secret" -o json "$KSQLDB_NAME"
+   ccloud ksql app create --cluster $CLUSTER --api-key "$KAFKA_API_KEY" --api-secret "$KAFKA_API_SECRET" -o json "$KSQLDB_NAME"
    ccloud api-key create --service-account $SERVICE_ACCOUNT_ID --resource $RESOURCE -o json    // for ksqlDB
 
    ccloud kafka acl create --allow --service-account $SERVICE_ACCOUNT_ID --operation <....>    // permissive ACLs for all services
@@ -119,7 +119,7 @@ Create a ccloud-stack
 
 #. ``ccloud-stack`` configures permissive ACLs with wildcards, which is useful for development and learning environments. In production, configure much stricter ACLs.
 
-   If you ran without ksqlDB (in the output below, service account ID is ``119612``):
+   If you ran without ksqlDB:
 
    .. code-block:: text
 
@@ -138,7 +138,7 @@ Create a ccloud-stack
         User:186607      | ALLOW      | WRITE            | TOPIC            | *             | LITERAL
         User:186607      | ALLOW      | READ             | TOPIC            | *             | LITERAL
 
-   If you ran with ksqlDB (in the output below, ksqlDB service account ID is ``119613``):
+   If you ran with ksqlDB:
 
    .. code-block:: text
 
--- a/ccloud/docs/ccloud-stack.rst
+++ b/ccloud/docs/ccloud-stack.rst
@@ -54,8 +54,8 @@ By default, the |ccloud| ksqlDB app is not created with ``ccloud-stack``, you ha
    ccloud api-key create --service-account $SERVICE_ACCOUNT_ID --resource $RESOURCE -o json    // for schema-registry
 
    # By default, ccloud-stack does not enable Confluent Cloud ksqlDB, but if you explicitly enable it:
-   ccloud ksql app create --cluster $CLUSTER -o json "$KSQLDB_NAME"
-   ccloud api-key create --service-account $KSQLDB_SERVICE_ACCOUNT_ID --resource $RESOURCE -o json    // for ksqlDB
+   ccloud ksql app create --cluster $CLUSTER --api-key "$kafka_api_key" --api-secret "$kafka_api_secret" -o json "$KSQLDB_NAME"
+   ccloud api-key create --service-account $SERVICE_ACCOUNT_ID --resource $RESOURCE -o json    // for ksqlDB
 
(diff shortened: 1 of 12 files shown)


utils/ccloud_library.sh
```diff
@@ -103,7 +105,7 @@ function ccloud::validate_version_ccloud_cli() {
 
   ccloud::validate_ccloud_cli_installed || exit 1
 
-  REQUIRED_CCLOUD_VER=${1:-"1.7.0"}
+  REQUIRED_CCLOUD_VER=${1:-$CCLOUD_MIN_VERSION}
```
ybyzek: @rspurgeon `REQUIRED_CCLOUD_VER` is treated like a minimum (see 2 lines below), so can this param be removed, and then use `CCLOUD_MIN_VERSION` instead below?

ccloud/docs/ccloud-stack.rst
```diff
@@ -54,8 +54,8 @@ By default, the |ccloud| ksqlDB app is not created with ``ccloud-stack``, you ha
    ccloud api-key create --service-account $SERVICE_ACCOUNT_ID --resource $RESOURCE -o json    // for schema-registry
 
    # By default, ccloud-stack does not enable Confluent Cloud ksqlDB, but if you explicitly enable it:
-   ccloud ksql app create --cluster $CLUSTER -o json "$KSQLDB_NAME"
-   ccloud api-key create --service-account $KSQLDB_SERVICE_ACCOUNT_ID --resource $RESOURCE -o json    // for ksqlDB
+   ccloud ksql app create --cluster $CLUSTER --api-key "$kafka_api_key" --api-secret "$kafka_api_secret" -o json "$KSQLDB_NAME"
```
ybyzek: This should use same parameter convention as the rest of the lines obove (capitalized, etc)

ccloud/docs/ccloud-stack.rst
```diff
@@ -70,7 +70,7 @@ Prerequisites
 =============
 
 - Create a user account in `Confluent Cloud <https://www.confluent.io/confluent-cloud/>`__
-- Local install of `Confluent Cloud CLI <https://docs.confluent.io/ccloud-cli/current/install.html>`__ v1.13.0 or later.
+- Local install of `Confluent Cloud CLI <https://docs.confluent.io/ccloud-cli/current/install.html>`__ v1.22.1 or later.
```
ybyzek, line 73: Excellent, thanks for updating the docs too

ccloud/docs/ccloud-stack.rst
```diff
@@ -122,73 +122,61 @@ Create a ccloud-stack
...
+        User:186607      | ALLOW      | DESCRIBE_CONFIGS | TOPIC            | *             | LITERAL
+        User:186607      | ALLOW      | WRITE            | TOPIC            | *             | LITERAL
+        User:186607      | ALLOW      | READ             | TOPIC            | *             | LITERAL
 
    If you ran with ksqlDB (in the output below, ksqlDB service account ID is ``119613``):
 
    .. code-block:: text
-   
-        ServiceAccountId | Permission |    Operation     |     Resource     |             Name             |   Type    
+
```
ybyzek, line 144: See line above, does it need to change (no ksqlDB SA)

>   If you ran with ksqlDB (in the output below, ksqlDB service account ID is ``119613``):

ccloud/docs/ccloud-stack.rst
```diff
@@ -317,7 +305,7 @@ If you don't want to create and destroy a ``ccloud-stack`` using the provided ba
 
    .. code:: bash
 
-      ccloud::destroy_ccloud_stack
+      ccloud::destroy_ccloud_stack $SERVICE_ACCOUNT_ID
```
ybyzek, line 308: Good catch

ybyzek (commented):
@rspurgeon left a few comments.  Can you please document which examples you have validated...beyond ccloud-stack...to ensure ksqlDB works?  e.g. cp-quickstart or others?

ccloud/docs/ccloud-stack.rst
rspurgeon: With this comment, I'm not certain if you're concerned with only the documentation or the code and the doc?  If you're only thinking about the documentation, and you're not concnered with the docs loosely matching the actual code (which it currently does), I suggest we rework this doc section to use sample values instead of variables, which aren't helpful to a user trying grok what commands are actually run. Or, we could more generalize this section to describe the types of behaviors (maybe with links to the actual ccloud docs) that the script performs, as I don't think this list encompasses everything required as it stands now.

WDYT?

ybyzek: Purely doc.  All lines above and below use capitalization.

```suggestion
   ccloud ksql app create --cluster $CLUSTER --api-key "$KAFKA_API_KEY" --api-secret "$KAFKA_API_SECRET" -o json "$KSQLDB_NAME"
```

utils/ccloud_library.sh
rspurgeon: Yes, I can do that. My hesitation is changing global variable assigments in a library because I don't exactly know how it is used by any client (outside this repo or maybe even ones we don't control that someone has built). Our wide use of global variables reduces my confidence in changing code without side effects, so I try to change as little as possible.

rspurgeon:
> @rspurgeon left a few comments. Can you please document which examples you have validated...beyond ccloud-stack...to ensure ksqlDB works? e.g. cp-quickstart or others?

Updated the PR description with this

ccloud/docs/ccloud-stack.rst
rspurgeon: Yes, thank you.  Will push a change with the other docs change

utils/ccloud_library.sh
ybyzek: 1. `REQUIRED_CCLOUD_VER` is not exported.
2. "There are no guarantees for backwards compatibility"

tl;dr Risk is low to break things unwittingly.  Let's optimize for reducing maintenance burden on ourselves.

rspurgeon: All variables in bash are global by default, export is not required for side effects.

rspurgeon:
@ybyzek Staging docs have been updated and suggestions addressed

ccloud/docs/ccloud-stack.rst
rspurgeon: staging docs have this change now

utils/ccloud_library.sh
ybyzek: Yeah, you're right.  I was thinking of the bash script use case, not function use case.

Regardless, no guarantees for backwards compatibility.  Let's please remove `REQUIRED_CCLOUD_VER`

rspurgeon: it's removed.  I'd like to see us transition to `local` so it's completely safe for function local changes.

ybyzek: @rspurgeon would `local` also address the issues encountered when running it back to back?

rspurgeon: I'm not 100% certain.  Some of our instructions rely on the user's shell having particular values set in order to perform demos ( example ccloud::destroy_ccloud_stack $SERVICE_ACCOUNT_ID).  I'm not sure where everything intersects with the issue of multiple runs and our usage of global vars.  I could certainly research and get back

ybyzek: I was asking just out of curiosity ;) to keep in my mind if/when we revisit those issues.  No need to research this now, but thank you for the idea in the first place!

ybyzek:
@rspurgeon LGTM, verified with microservices-orders

ybyzek (approved):
Thank you @rspurgeon 🎉

PR #951, opened 2021-02-26
remove_timeout_req from ccloud-observability

ybyzek:
### Description 

@awalther28 please confirm but I think `timeout` and `mvn` are not required on the user's host

### Author Validation

- [ ] Documentation

### Reviewer Tasks


awalther28 (approved):
You are correct maven and timeout are no longer requirements.

PR #953, opened 2021-03-01
DOCS-6760: Add CCloud promo code to connect-streams-pipeline and clic…

ybyzek:
…kstream demos

### Description 

https://confluentinc.atlassian.net/browse/DOCS-6760

### Author Validation

- [x] Documentation -- I validated it builds.  So reviewers can just comment on text

### Reviewer Tasks

### DIFF
--- a/clickstream/docs/index.rst
+++ b/clickstream/docs/index.rst
@@ -45,6 +45,7 @@ are quite large and depending on your network connection may take
 
        ./start.sh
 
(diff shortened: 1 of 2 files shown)


awalther28 (approved):
This looks good to me and appears to complete the objective of the ticket. 
It would be nice if we had a fully managed connector example to point people to here. The only one I can think of is buried in beginner cli ccloud. I think this text suffices since there is a goal to migrate the content to CCloud.
Note, I did not build the docs. I only reviewed the text.

PR #957, opened 2021-03-09
Add --prompt to ccloud::login_ccloud_cli() function

gianlucanatali:
### Description 

ccloud::login_ccloud_cli() function accept username and password, since now ccloud CLI support saved credentials, the --prompt flag is needed to force the cli to prompt for the password (in case creds are saved)

### Author Validation

@ybyzek If I understand correctly this function is not used much...

### Reviewer Tasks

Not sure if any sample uses this function...


ybyzek (approved):
@gianlucanatali LGTM , thank you for the PR.

Issue #969, opened 2021-04-01
ccloud-observability fails to load java dependencies


awalther28:
**Description**
ccloud-observability

Producer won't start due to an error resolving the `confluent.version`.

```
localhost:ccloud-observability awalther$ mvn clean install -U
[INFO] Scanning for projects...
Downloading from confluent-artifactory-central: https://confluent.jfrog.io/confluent/maven-public/joda-time/joda-time/maven-metadata.xml
Downloaded from confluent-artifactory-central: https://confluent.jfrog.io/confluent/maven-public/joda-time/joda-time/maven-metadata.xml (1.7 kB at 3.2 kB/s)
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for io.confluent:kafka-avro-serializer:jar must be a valid version but is '${confluent.version}'. @ line 58, column 22
[ERROR] 'dependencies.dependency.version' for io.confluent:kafka-streams-avro-serde:jar must be a valid version but is '${confluent.version}'. @ line 63, column 22
[ERROR] 'dependencies.dependency.version' for io.confluent:kafka-json-serializer:jar must be a valid version but is '${confluent.version}'. @ line 83, column 22
[ERROR] 'build.plugins.plugin.version' for io.confluent:kafka-schema-registry-maven-plugin must be a valid version but is '${confluent.version}'. @ line 157, column 26
 @ 
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]   
[ERROR]   The project io.confluent:clients-example:6.1.1 (/Users/awalther/Documents/devx/examples/ccloud-observability/pom.xml) has 4 errors
[ERROR]     'dependencies.dependency.version' for io.confluent:kafka-avro-serializer:jar must be a valid version but is '${confluent.version}'. @ line 58, column 22
[ERROR]     'dependencies.dependency.version' for io.confluent:kafka-streams-avro-serde:jar must be a valid version but is '${confluent.version}'. @ line 63, column 22
[ERROR]     'dependencies.dependency.version' for io.confluent:kafka-json-serializer:jar must be a valid version but is '${confluent.version}'. @ line 83, column 22
[ERROR]     'build.plugins.plugin.version' for io.confluent:kafka-schema-registry-maven-plugin must be a valid version but is '${confluent.version}'. @ line 157, column 26
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[...]

PR #970, opened 2021-04-01
GH-969: ccloud-observability fails to load java dependencies

awalther28:
### Description 
https://github.com/confluentinc/examples/issues/969

This PR fixes the confluent.version variable used by the pom.xml in the ccloud-observability project. As it stands rn, the variable will not resolve to a version number. Looking through the parent pom files, it doesn't appear that this variable is defined, rather a similarly named variable is being used "confluent.version.range".  See https://github.com/confluentinc/common/blob/580cfe0ddb94ad306c43218b6ee07a1c8a9bb324/pom.xml#L133. 

### Author Validation

 - [x] ccloud-observability

### Reviewer Tasks

### DIFF
--- a/ccloud-observability/pom.xml
+++ b/ccloud-observability/pom.xml
@@ -55,12 +55,12 @@
         <dependency>
             <groupId>io.confluent</groupId>
             <artifactId>kafka-avro-serializer</artifactId>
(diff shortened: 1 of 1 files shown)


ybyzek (approved):
Approving for expedited merge

PR #976, opened 2021-04-19
multi-datacenter: update the max wait time for connect cluster to start

nsharma-git:
### Description 

This change is for example demo for multi-datacenter. At times the the check for cluster status fails because it takes more than 120 seconds ( default time out ). 

```
Waiting up to 120 seconds for Connect to start
........................ERROR: Failed after 120 seconds. Please troubleshoot and run again.
```

Increasing the timeout to make it more robust across different docker environments. 

### Author Validation

[ ] multi-datacenter

### DIFF
--- a/multi-datacenter/start.sh
+++ b/multi-datacenter/start.sh
@@ -8,7 +8,7 @@ check_jq || exit 1
 docker-compose up -d
 
 # Verify Kafka Connect workers have started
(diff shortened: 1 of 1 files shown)


ybyzek (approved):
@nsharma-git Thank you 🎉 LGTM.

PR #983, opened 2021-04-29
Update docker-compose.yml

mcascallares:
Automatic observer promotion requires version:2 for log.placement.constraints

### Description 

### Author Validation

### Reviewer Tasks

### DIFF
--- a/multiregion/docker-compose.yml
+++ b/multiregion/docker-compose.yml
@@ -88,7 +88,7 @@ services:
       KAFKA_CONFLUENT_METRICS_REPORTER_BOOTSTRAP_SERVERS: broker-ccc:19098
       KAFKA_CONFLUENT_MONITORING_INTERCEPTER_BOOTSTRAP_SERVERS: broker-ccc:19098
       KAFKA_JMX_PORT: 8091
-      KAFKA_CONFLUENT_LOG_PLACEMENT_CONSTRAINTS: '{"version": 1,"replicas": [{"count": 2, "constraints": {"rack": "west"}}], "observers": [{"count": 2, "constraints": {"rack": "east"}}]}'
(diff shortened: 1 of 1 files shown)


ybyzek:
@mcascallares thanks for the PR.  Can you please update the description with details and reference links if available?  For example, the tutorial appears to work today without this PR -- is that actually not true?

mcascallares:
Hi @ybyzek 

We just tested it today with version 1, simulating the failover scenario and it didn't work. 

@csarrazi discovered this blog post https://www.confluent.io/blog/automatic-observer-promotion-for-safe-multi-datacenter-failover-in-confluent-6-1/ that explicitly states "Automatic Observer Promotion is activated by changing the version of the replica placement policy to two."

Hope this helps!

ybyzek:
> We just tested it today with version 1, simulating the failover scenario and it didn't work.

Can you please elaborate on "it didn't work"?  What was the symptom?  This will help me validate the PR

mcascallares:
Sorry, my bad. 

"It didn't work" -> when we brought down one of the ISR, and the observerPromotionPolicy was met, it didn't promote one of the observers. @csarrazi discovered that blog post, we switched from v1 to v2 as my PR, and the automatic promotion was triggered.

It is also highlighted in our docs https://docs.confluent.io/platform/current/multi-dc-deployments/multi-region.html#automatic-observer-promotion

ybyzek:
There are specific topics that showcase the automatic observer promotion behavior, and they all use version=2 (see directory https://github.com/confluentinc/examples/tree/6.1.1-post/multiregion/config):

https://github.com/confluentinc/examples/blob/6.1.1-post/multiregion/config/placement-multi-region-async-op-leader-is-observer.json#L2
https://github.com/confluentinc/examples/blob/6.1.1-post/multiregion/config/placement-multi-region-async-op-under-min-isr.json#L2
https://github.com/confluentinc/examples/blob/6.1.1-post/multiregion/config/placement-multi-region-async-op-under-replicated.json#L2

So the PR as it stands today -- I'm still not clear what affect it has in the tutorial. 

cc: @thomaskwscott who added Automatic Observer Promotion into this tutorial, for additional comment

PR #984, opened 2021-04-29
DEVX-2589: Update Confluent CLI version min/max for CP 6.2

ybyzek:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2589

1. Main issue: CP versions pinned to specific CLI versions
2. Side issue: reading CP version the first time CLI is run, account for update check

### Author Validation

- [x] cp-quickstart (started but cannot fully run do to another issue)

### Reviewer Tasks

### DIFF
--- a/utils/helper.sh
+++ b/utils/helper.sh
@@ -52,8 +52,8 @@ function validate_version_confluent_cli_for_cp() {
 
   validate_version_confluent_cli_v2 || exit 1
 
-  VER_MIN="1.11.0"
-  VER_MAX="1.22.0"
+  VER_MIN="1.26.0"
+  VER_MAX="1.26.1"
   CLI_VER=$(get_version_confluent_cli)
 
   if version_gt $VER_MIN $CLI_VER || version_gt $CLI_VER $VER_MAX ; then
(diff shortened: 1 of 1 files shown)


utils/helper.sh
```diff
@@ -52,8 +52,8 @@ function validate_version_confluent_cli_for_cp() {
 
   validate_version_confluent_cli_v2 || exit 1
 
-  VER_MIN="1.11.0"
-  VER_MAX="1.22.0"
+  VER_MIN="1.26.0"
+  VER_MAX="1.26.1"
```
ybyzek, line 56: Since we don't know max at this time, this hedges a bit.

ethanruhe: There is definitely a 1.27.0. I was trying to think of a clever way to automate updating this based on the results from `curl -sL --http1.1 https://cnfl.io/cli | sh -s -- -l`, but this seems like maybe more trouble than it's worth.

ybyzek: We explicitly do NOT want to curl because when 7.0 is released with a new and possibly incompatible CLI, 6.2.x should NOT use it.

`VER_MAX` should be the max compatible with 6.2.x

ybyzek:
To unblock QA, I'm going to merge this PR now.  We can reset MAX later if needed.

PR #1011, opened 2021-07-26
DEVX-2648 Wrong config file variable name at kafkacat-example

RamanVerma:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2648

Fix for the `kafkacat-example.sh` file. The config file specification was wrong.

### Author Validation

- [x] clients/cloud Tested on a ccloud cluster

### DIFF
--- a/clients/cloud/kafkacat/kafkacat-example.sh
+++ b/clients/cloud/kafkacat/kafkacat-example.sh
@@ -22,11 +22,11 @@ kafka-topics --bootstrap-server `grep "^\s*bootstrap.server" $CONFIG_FILE | tail
 # Produce messages
 num_messages=10
 (for i in `seq 1 $num_messages`; do echo "alice,{\"count\":${i}}" ; done) | \
-   kafkacat -F $HOME/.confluent/config \
+   kafkacat -F $CONFIG_FILE \
             -K , \
             -P -t $topic_name
 
 # Consume messages
-kafkacat -F $HOME/.confluent/config \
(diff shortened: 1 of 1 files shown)


ybyzek (commented):
@RamanVerma internal tooling automatically bumps versions for new releases, in the appropriate branches.  This should not be manually done.

RamanVerma:
@ybyzek hmm, some thing went wrong with my local branch. I see that there are 14 files changed in this PR. I wanted to change just the kafkacat-example. I will correct it and remove the unwanted changes in some time. Thanks for looking into it.

RamanVerma:
Also @ybyzek, why do we have 6.2.0-post as the default branch. Just wanted to make sure that is the correct default (and not master)

ybyzek (changes requested):
This actually won't work because the topic creation argument expects a Java-based config

```
sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required username='<...>' password='<...>';
```

whereas kafkacat expects the librdkafka version of this:

```
sasl.username=<...>
sasl.password=<...>
```

So this PR won't work as-is.  A broader rewrite of the script and docs for the kafkacat example needs to be done.

You can do it in this PR or we can leave the jira DEVX-2648 open to address this.

ybyzek:
This has been resolved and documentation for running this example is live at https://docs.confluent.io/platform/current/tutorials/examples/clients/docs/kcat.html

PR #1023, opened 2021-08-31
Rename Confluent Cloud UI to Confluent Cloud Console

ybyzek:
### Description 

### Author Validation

- [x] Documentation -- I validated docs build locally

### Reviewer Tasks

### DIFF
--- a/ccloud-observability/docs/consumer-scenarios/increasing-consumer-lag.rst
+++ b/ccloud-observability/docs/consumer-scenarios/increasing-consumer-lag.rst
@@ -70,7 +70,7 @@ Diagnose the problem
       If a client is `properly tuned <https://docs.confluent.io/cloud/current/client-apps/optimizing/index.html>`__ and has adequate resources
       an increase in throughput metrics or fetch metrics won't necessarily mean the consumer lag will increase.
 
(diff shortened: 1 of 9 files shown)


rspurgeon (approved):
LGTM

vsyu:
For the URLs, are we using https://confluent.cloud/login because the user likely already has an account? https://www.confluent.io/confluent-cloud/tryfree/ is generally the preferred link, but understandably in some cases it may not be appropriate. 

I checked with Leslie on the shorthand for "Confluent Cloud Console" as well. She shared: "The Docs rule is that the first time in blog/ page we should use the full name 'Confluent Cloud Console' and the we can refer to it as 'Cloud Console' for an abbreviated format.

ybyzek:
@vsyu thanks for the feedback, and I agree.  I have updated links as needed in the latest commit.

vsyu:
Thanks for the updates, @ybyzek! The only other suggestion would be to update line 225 to point to https://www.confluent.io/confluent-cloud/tryfree/ instead of https://confluent.cloud/.

ybyzek:
@vsyu that line is intentionally pointing to https://confluent.cloud/.  The user by this point has already been shown the `tryfree` link and the intention is to send them now to the console itself :)

vsyu:
@ybyzek ah, good to know! Makes sense :)

PR #1034, opened 2021-09-21
DEVX-2673: set env before ccloud destroy

ybyzek:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2673

Set env before ccloud-stack destroy.
This handles the use case where user switched over to another environment in between runs (or logged out of ccloud CLI), either way, the proper environment would not have been set.

### Author Validation

- [x] cp-quickstart

### Reviewer Tasks

### DIFF
--- a/utils/ccloud_library.sh
+++ b/utils/ccloud_library.sh
@@ -1035,7 +1035,8 @@ function ccloud::destroy_ccloud_stack() {
   fi
 
   SERVICE_ACCOUNT_ID=$1
-  ENVIRONMENT=$(ccloud::get_environment_id_from_service_id $SERVICE_ACCOUNT_ID)
+  ENVIRONMENT=${ENVIRONMENT:-$(ccloud::get_environment_id_from_service_id $SERVICE_ACCOUNT_ID)}
+
   ccloud environment use $ENVIRONMENT || exit 1
 
   PRESERVE_ENVIRONMENT="${PRESERVE_ENVIRONMENT:-false}"
--- a/utils/ccloud_library.sh
+++ b/utils/ccloud_library.sh
@@ -1035,6 +1035,8 @@ function ccloud::destroy_ccloud_stack() {
   fi
 
   SERVICE_ACCOUNT_ID=$1
+  ENVIRONMENT=$(ccloud::get_environment_id_from_service_id $SERVICE_ACCOUNT_ID)
+  ccloud environment use $ENVIRONMENT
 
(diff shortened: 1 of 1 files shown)


utils/ccloud_library.sh
```diff
@@ -1035,6 +1035,8 @@ function ccloud::destroy_ccloud_stack() {
   fi
 
   SERVICE_ACCOUNT_ID=$1
+  ENVIRONMENT=$(ccloud::get_environment_id_from_service_id $SERVICE_ACCOUNT_ID)
+  ccloud environment use $ENVIRONMENT
```
rspurgeon: @ybyzek I've been worried about the destroy script as well w/ statefulnees of the `ccloud` CLI, so this is a good change.  I'm wondering does the demo fail fast if the `ccloud environment use` fails here so that any further commands do no damange?

ybyzek: @rspurgeon to address this, latest commit now fails fast with `|| exit 1`.

rspurgeon (approved):
LGTM

javabrett (changes requested):
My test failed, but I know why.

I always re-use the same `ENVIRONMENT`, which I have set permanently in my shell env - so it is not named via the service account ID convention, so can't be found during the clean-up either.

This works during the create phase as there is a `-z` check on it before setting:

https://github.com/confluentinc/examples/blob/e722e1de9bae0367fa487bce0a0ccbc70be78338/utils/ccloud_library.sh#L933-L941

So we could do similar here, or use an `ENVIRONMENT=${ENVIRONMENT:-...}` style default in order to preference from my environment `ENVIRONMENT` :).

javabrett:
With this change:

```
@@ -1035,7 +1037,7 @@ function ccloud::destroy_ccloud_stack() {
   fi

   SERVICE_ACCOUNT_ID=$1
-  ENVIRONMENT=$(ccloud::get_environment_id_from_service_id $SERVICE_ACCOUNT_ID)
+  ENVIRONMENT=${ENVIRONMENT:-$(ccloud::get_environment_id_from_service_id $SERVICE_ACCOUNT_ID)}
   ccloud environment use $ENVIRONMENT || exit 1

   PRESERVE_ENVIRONMENT="${PRESERVE_ENVIRONMENT:-false}"
   ```
   
   ... tear-down is successful - I did not re-test the standard, `ENVIRONMENT` not pre-set in env path.

ybyzek:
@javabrett please see latest commit, I believe it addresses both codepaths now.

javabrett (approved):

PR #1046, opened 2021-11-10
replaced Serialized to Grouped

ldai100:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2692

Serialized class was deprecated. Maven build will fail. It was replaced by Grouped class.

### Author Validation

No documentation change is required. Same code compiled and ran according to confluent doc. It ran successful.

### Reviewer Tasks


rspurgeon (approved):
LGTM

PR #1068, opened 2022-01-30
Avoid the cluster use command to write to standard outptut because it

bleporini:
breaks the result of the function.

### Description 

### Author Validation

### Reviewer Tasks


ybyzek:
@bleporini in my environment, the script works without this PR, so I'm curious what's different when run in your environment?

ybyzek (commented):
@bleporini from our Slack convo, a more efficient way to go may be https://confluentinc.atlassian.net/browse/DEVX-2713 which would sidestep the issue completely (which, btw, still not clear why this isn't causing errors in my env but I understand why it could be a problem potentially)

UPDATE: can this PR be closed in favor of https://github.com/confluentinc/examples/pull/1069 ?

javabrett:
@bleporini do you still experience the issues this PR aimed to address, or can it be closed?

bleporini:
@javabrett considering the changes that have been made to this function since then, it's not relevant anymore, you can close it.

PR #1070, opened 2022-02-07
DEVX-2714: update Confluent CLI syntax and checking ids

ybyzek:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2714

- update CLI change for ksql cluster
- Better to use json with jq over grep

### Author Validation

### Reviewer Tasks

### DIFF
--- a/cloud-etl/create_ksqldb_app.sh
+++ b/cloud-etl/create_ksqldb_app.sh
@@ -27,8 +27,8 @@ ccloud::validate_credentials_ksqldb "$KSQLDB_ENDPOINT" "$CONFIG_FILE" "$KSQLDB_B
 echo -e "Create output topics $KAFKA_TOPIC_NAME_OUT1 and $KAFKA_TOPIC_NAME_OUT2, and ACLs to allow the ksqlDB application to run\n"
 confluent kafka topic create $KAFKA_TOPIC_NAME_OUT1
 confluent kafka topic create $KAFKA_TOPIC_NAME_OUT2
-ksqlDBAppId=$(confluent ksql app list | grep "$KSQLDB_ENDPOINT" | awk '{print $1}')
-confluent ksql app configure-acls $ksqlDBAppId $KAFKA_TOPIC_NAME_IN $KAFKA_TOPIC_NAME_OUT1 $KAFKA_TOPIC_NAME_OUT2
+ksqlDBAppId=$(confluent ksql cluster list | grep "$KSQLDB_ENDPOINT" | awk '{print $1}')
+confluent ksql cluster configure-acls $ksqlDBAppId $KAFKA_TOPIC_NAME_IN $KAFKA_TOPIC_NAME_OUT1 $KAFKA_TOPIC_NAME_OUT2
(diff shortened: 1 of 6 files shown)


cloud-etl/create_ksqldb_app.sh
```diff
@@ -27,8 +27,8 @@ ccloud::validate_credentials_ksqldb "$KSQLDB_ENDPOINT" "$CONFIG_FILE" "$KSQLDB_B
 echo -e "Create output topics $KAFKA_TOPIC_NAME_OUT1 and $KAFKA_TOPIC_NAME_OUT2, and ACLs to allow the ksqlDB application to run\n"
 confluent kafka topic create $KAFKA_TOPIC_NAME_OUT1
 confluent kafka topic create $KAFKA_TOPIC_NAME_OUT2
-ksqlDBAppId=$(confluent ksql app list | grep "$KSQLDB_ENDPOINT" | awk '{print $1}')
-confluent ksql app configure-acls $ksqlDBAppId $KAFKA_TOPIC_NAME_IN $KAFKA_TOPIC_NAME_OUT1 $KAFKA_TOPIC_NAME_OUT2
+ksqlDBAppId=$(confluent ksql cluster list | grep "$KSQLDB_ENDPOINT" | awk '{print $1}')
```
rspurgeon: Would it make sense to convert this to using `json` output as well?

ybyzek: @rspurgeon good catch.  Updated in latest commit

cp-quickstart/start-cloud.sh
```diff
@@ -93,13 +93,13 @@ retry $MAX_WAIT ccloud::validate_ccloud_ksqldb_endpoint_ready $KSQLDB_ENDPOINT |
 print_pass "Confluent Cloud KSQL is UP"
 
 printf "Obtaining the ksqlDB App Id\n"
-CMD="confluent ksql app list -o json | jq -r '.[].id'"
+CMD="confluent ksql cluster list -o json | jq -r '.[].id'"
```
rspurgeon, line 96: FYC: Now that we're using json, we can be more reliable with these commands. The array might contain more than one ksqldb app, and then we are storing the wrong app id and maybe causing issues for the user.  Instead we could be specific with `jq`...

```
confluent ksql cluster list -o json | jq --arg KSQLDB_ENDPOINT "$KSQLDB_ENDPOINT" -r '.[] | select(.endpoint==$KSQLDB_ENDPOINT) | .id'
```

ybyzek: Point take.  In this case: if a user is running this script `cp-quickstart/start-cloud.sh`, it's a new environment so there is only one ksqlDB app.

rspurgeon: Yes, unless the user is running other commands during long scripting runs (like while the app provisions) and then results in confused state in the CLI.

rspurgeon (approved):
lgtm with minor comments. Verified the beginner-cloud and cp-quickstart (didn't run etl).

ybyzek:
Thanks for the review @rspurgeon

PR #1077, opened 2022-02-22
DEVX-2719: update Confluent CLI name and syntax for CP 7.1.0

ybyzek:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2719

CLI syntax and name changes

### Author Validation

- [x] security/rbac

### Reviewer Tasks

- [ ] security/rbac


utils/helper.sh
```diff
@@ -54,8 +47,8 @@ function version_gt() {
 function validate_version_confluent_cli_for_cp() {
   validate_version_confluent_cli_v2 || exit 1
 
-  VER_MIN="1.40.0"
-  VER_MAX="1.43.1"
+  VER_MIN="2.3.1"
```
ybyzek, line 50: @DABH @ethanruhe do we know what versions of CLI CP 7.1.0 will ship with?  (min/max)

security/rbac/docs/index.rst
```diff
@@ -146,11 +146,11 @@ Broker
    .. code:: bash
    
       # Broker Admin
-      confluent-v1 iam rolebinding create --principal User:$USER_ADMIN_SYSTEM --role SystemAdmin --kafka-cluster-id $KAFKA_CLUSTER_ID
+      confluent iam rbac create --principal User:$USER_ADMIN_SYSTEM --role SystemAdmin --kafka-cluster-id $KAFKA_CLUSTER_ID
```
brianstrauch: https://docs.confluent.io/confluent-cli/current/command-reference/iam/rbac/role-binding/confluent_iam_rbac_role-binding_create.html#confluent-iam-rbac-role-binding-create
```suggestion
      confluent iam rbac role-binding create --principal User:$USER_ADMIN_SYSTEM --role SystemAdmin --kafka-cluster-id $KAFKA_CLUSTER_ID
```

utils/helper.sh
```diff
@@ -82,8 +75,8 @@ function check_python() {
 }
 
 function check_confluent_binary() {
-  if [[ ! -f ~/confluent-v1/confluent ]]; then
-    echo "The alias 'confluent-v1' is not set up. Follow the migration guide (https://docs.confluent.io/confluent-cli/current/migrate.html#run-multiple-clis-in-parallel) and try again."
+  if [[ ! -f ~/confluent/confluent ]]; then
+    echo "The alias 'confluent' is not set up. Use the Confluent CLI bundled with Confluent Platform and try again."
```
brianstrauch: We're no longer using an alias here, so let's revert to the original logic.
```suggestion
  if [[ $(type confluent 2>&1) =~ "not found" ]]; then
    echo "'confluent' is not found. Install Confluent Platform if you want to use Confluent CLI."
```

brianstrauch (changes requested):

utils/helper.sh
ybyzek: Ack and done in latest commit

security/rbac/docs/index.rst
ybyzek: Ack and done in latest commit

brianstrauch (approved):

utils/helper.sh
rspurgeon: Not official of course, but this is reported on my current setup:
```
➜ $CONFLUENT_HOME/bin/confluent version
confluent - Confluent CLI

Version:     v2.3.1
Git Ref:     bc90de55
Build Date:  2021-12-10T21:02:09Z
Go Version:  go1.16.3 (darwin/amd64)
Development: false
```

rspurgeon (approved):
lgtm, i verified the scripted steps

utils/helper.sh
ethanruhe: It'll ship with 2.6.0 and CP will be officially interoperable with "2.6.0 through the most recent version." 

It is a little odd to not have 2.0.0 compatible with any CP version, but we know the early 2.x releases had some meaningful bugs. Those were cleaned up in the intervening releases, so I think pinning the earliest officially interoperable version to the one that comes packaged is the best thing for customers.

brianstrauch: Actually, [2.6.1](https://github.com/confluentinc/cli/releases/tag/v2.6.1)

ybyzek: Thanks!  I'm leaving at 2.3.1 min for now since that is what is shipped with the nightlies.

aakashenv (approved):
Hello @ybyzek, Thanks for the fix. For future purpose: I am part of QE team however, I am not responsible for PR validation and it can be done by eng team. I have validated this time and was able to bring up the cluster successfully without any of the error mentioned in the bug.

PR #1079, opened 2022-02-27
DEVX-2723: call confluent kafka cluster use before stack tear-down

javabrett:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2723

Specifies `--cluster` ID when running `confluent kafka acl delete` during tear-down, to prevent attempt to delete ACL from the context cluster, should a cluster different from the demo cluster be selected when `stop-cloud.sh` is run.

### Author Validation

 - [X] ccloud/ccloud-stack

### Reviewer Tasks

### DIFF
--- a/utils/ccloud_library.sh
+++ b/utils/ccloud_library.sh
@@ -1057,6 +1057,7 @@ function ccloud::destroy_ccloud_stack() {
   echo "Destroying Confluent Cloud stack associated to service account id $SERVICE_ACCOUNT_ID"
 
   local cluster_id=$(confluent kafka cluster list -o json | jq -r 'map(select(.name == "'"$CLUSTER_NAME"'")) | .[].id')
+  confluent kafka cluster use $cluster_id 2>/dev/null
 
   # Delete associated ACLs
   ccloud::delete_acls_ccloud_stack $cluster_id $SERVICE_ACCOUNT_ID
--- a/utils/ccloud_library.sh
+++ b/utils/ccloud_library.sh
@@ -1068,7 +1068,7 @@ function ccloud::destroy_ccloud_stack() {
   fi
 
   # Delete connectors associated to this Kafka cluster, otherwise cluster deletion fails
-  confluent connect list --cluster $cluster_id -o json | jq -r '.[].id' | xargs -I{} confluent connect delete {}
+  confluent connect list --cluster $cluster_id -o json | jq -r '.[].id' | xargs -I{} confluent connect delete --cluster $cluster_id {}
 
   echo "Deleting CLUSTER: $CLUSTER_NAME : $cluster_id"
   confluent kafka cluster delete $cluster_id &> "$REDIRECT_TO"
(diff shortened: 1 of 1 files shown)


ybyzek:
Per Slack, instead of doing this PR, the fix could be to `use` the Kafka cluster ID identified

javabrett:
Added two commits to:
- Address DEVX-2724 also, by adding `--cluster` to `confluent connect delete` calls during tear-down
- Added call to `confluent kafka cluster use` prior to teardown: good hygiene to ensure the correct cluster is selected prior to tear-down.

utils/ccloud_library.sh
```diff
@@ -1055,8 +1056,11 @@ function ccloud::destroy_ccloud_stack() {
 
   echo "Destroying Confluent Cloud stack associated to service account id $SERVICE_ACCOUNT_ID"
 
+  local cluster_id=$(confluent kafka cluster list -o json | jq -r 'map(select(.name == "'"$CLUSTER_NAME"'")) | .[].id')
+  confluent kafka cluster use $cluster_id 2>/dev/null
```
ybyzek, line 1059: Given this statement, is it possible that the other `--cluster $CLUSTER_ID` are extraneous and can be removed?

javabrett: I decided both `--cluster` instances were safest and correct to retain, albeit unnecessary following successful `confluent kafka cluster use`.  Ideally we don't rely on this context for any commands - ideally all the commands are independent of context by fully specifying context via args.

But no, they aren't necessary, if removed I expect the bug would remain fixed.

ybyzek: For simplicity (and therefore maintenance0, my vote is to keep this PR to just these 2 lines here

```
  local cluster_id=$(confluent kafka cluster list -o json | jq -r 'map(select(.name == "'"$CLUSTER_NAME"'")) | .[].id')
   confluent kafka cluster use $cluster_id 2>/dev/null
```

And remove the redundancy of the other lines in this PR (`--cluster $CLUSTER_ID`)

javabrett: @ybyzek how would you feel about instead retaining the `--cluster` args over the `cluster use`?

That way there is no ambiguity in the commands, they are explicit and precise about what they select for deletion - if for example a user were able to fluke an accidental concurrent `confluent` command use in another shell, these deletes would remain correct and work even if the `cluster use` has been undone, or otherwise failed.

ybyzek: I understand the goal but there are so many other places in this ccloud library (aside from this function) where concurrent `confluent` commands can result in problems, that IMO it's safer to not set that expectation with the user.  It could be a separate effort to harden those functions to meet that expectation.  For now, given what little cycles we have for `ccloud-stack` in general, IMO it's better to optimize for simpler code.

davetroiano (approved):
LGTM!

PR #1097, opened 2022-04-22
DEVX-2740: BYOC from env parameters

ybyzek:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2740

2nd PR meant to address/obviate https://github.com/confluentinc/examples/pull/1094

### Author Validation

### Reviewer Tasks


utils/ccloud_library.sh
```diff
@@ -932,6 +944,25 @@ function ccloud::set_kafka_cluster_use() {
 }
 
 
+function ccloud::maybe_create_and_use_environment() {
+
+  ENVIORNMENT_NAME=$1
```
brianstrauch:
```suggestion
  ENVIRONMENT_NAME=$1
```

brianstrauch (approved):
CLI commands look good :+1:

utils/ccloud_library.sh
ybyzek: Thanks for catching this!  It was actually misspelled in TWO places, which is why it worked when I tested it.  Fixed in latest commit.

PR #1099, opened 2022-05-15
Fixed Exception caused by Empty Lines in config file.

JBeamer92:
### Description 

The config file provided in the C# setup contains empty lines with a Length of 0. This was causing Exceptions when trying to read the config file in the LoadConfig method because the empty lines were not being filtered out by the Where clause. By filtering out empty lines with a length of 0, the LoadConfig method completes without exceptions and the application completes as expected.

### Author Validation

Verified using .NET 6.0. Needs to be tested against earlier versions of .NET to prove backwards compatibility.

[ ] clients/cloud

### Reviewer Tasks

Needs to be tested against earlier versions of .NET to prove backwards compatibility.

[ ] clients/cloud

### DIFF
--- a/clients/cloud/csharp/Program.cs
+++ b/clients/cloud/csharp/Program.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2020 Confluent Inc.
+// Copyright 2020 Confluent Inc.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


ybyzek:
@bbejeck do you happen to have the right environment set up to validate this PR?

bbejeck:
> do you happen to have the right environment set up to validate this PR?

Yes - I have .NET versions 3, 5, and 6 installed - I can test against all 3

bbejeck (approved):

bbejeck:
Tested against .NET 6 and 5 - .NET 3 is pretty old and not compatible with `Confluent.Kafka 1.8.2` so IMHO going with version 5 is good enough

ybyzek:
@bbejeck thanks so much for validating and @JBeamer92 thanks again for the PR!  Merging now

PR #1105, opened 2022-05-31
Cloud events json schema

1123:
### Description 

This pull request adds an example for writing data respecting the Cloud Events Standard to Confluent Cloud, and registering the corresponding schema with Confluent Schema Registry. It is a somewhat minimal example for this requirement and independent of the other client samples in this repository. 

### Author Validation

[x] Documentation: This sample comes with its own README.md file. Not sure if it needs to be documented elsewhere. 
[x] clients/cloud: This is the directory where the sample has been added. It could be moved to another directory if needed. It should work just as well with an on-premise cluster. 

This pull request does not touch any of the other sample programs, and thus should not affect their functionality. 

### Reviewer Tasks


ybyzek:
@1123 thanks for the PR.  

1. The examples in the `clients/cloud/` folder are for languages, not specifications.  A better location may be to put this in `clients/` in a new subfolder called `cloudevents/`?
2. Should there be a consumer example?
3. If there isn't already, perhaps there should be a link added from docs.confluent.io
4. Please recommend another reviewer who is familiar with CloudEvents, who can review this PR from that perspective.

PR #1114, opened 2022-07-22
Avoid override file being pulled in

### LINKED ISSUE #1113: `.service "ksqldb-cli" has neither an image nor a build context specified: invalid compose project`
**Description**
This issue is encountered when trying to run `cp-quickstart/start-docker-kraft.sh`.  Error:

```
.service "ksqldb-cli" has neither an image nor a build context specified: invalid compose project
```

The temporary workaround is to remove `docker-compose.override.yml` in that folder because it is used by the other quickstarts but not the kraft one.  Need to think about long term fix

Reproducible by @ybyzek  and a colleague

ybyzek:
### Description 

See #1113 for details.

The way this PR addresses the problem is by being explicit about specifying the `docker-compose.yml` file, and that file only.  It avoids pulling in the override file (effectively ignores it), but it can't be deleted because it is used by other quickstarts in this folder.  Those use ksqlDB, whereas the KRaft one currently does not.

Note: this is a temporary quick fix to suppress the error.  The longer term fix is https://github.com/confluentinc/cp-all-in-one/issues/101 which is to get ksqlDB into the cp-all-in-one-kraft Docker Compose at https://github.com/confluentinc/cp-all-in-one/blob/7.2.1-post/cp-all-in-one-kraft/docker-compose.yml

### DIFF
--- a/cp-quickstart/start-docker-kraft.sh
+++ b/cp-quickstart/start-docker-kraft.sh
@@ -7,6 +7,7 @@ wget -O docker-compose.yml https://raw.githubusercontent.com/confluentinc/cp-all
 wget -O update_run.sh https://raw.githubusercontent.com/confluentinc/cp-all-in-one/${CONFLUENT_RELEASE_TAG_OR_BRANCH}/cp-all-in-one-kraft/update_run.sh
 chmod 744 update_run.sh
 
+export COMPOSE_FILE=docker-compose.yml
 ./stop-docker.sh
 
 docker-compose up -d


davetroiano (approved):
LGTM!

ybyzek:
cc: @davetroiano 

I'm working on adding ksqlDB to cp-all-in-one-kraft so this may be obsolete... For tracking: https://github.com/confluentinc/examples/pull/1117/files
https://github.com/confluentinc/cp-all-in-one/pull/104

ybyzek:
https://github.com/confluentinc/cp-all-in-one/pull/104 has now been merged so this fix is no longer needed.

PR #1121, opened 2022-08-06
Corrected Infinite for loop

abhilash-od:
### Description 

### Author Validation

### Reviewer Tasks

### DIFF
--- a/clients/cloud/go/consumer.go
+++ b/clients/cloud/go/consumer.go
@@ -66,8 +66,8 @@ func main() {
 
 	// Process messages
 	totalCount := 0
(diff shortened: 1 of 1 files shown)


davetroiano:
@abhilash-od I don't follow this change. With it, the consumer _will_ loop infinitely and ignore the user's interrupt -- `run` is set to `false` on interrupt so we want the `for run == true` condition in order to break out of the loop and terminate. Given this, I'm going to close out this PR. Feel free to reply with more details on the infinite loop behavior you're seeing and change you're looking to make

PR #1130, opened 2022-10-04
ccloud-observability: Fix JMX client metrics for Docker Compose 2.x

javabrett:
### Description 

https://confluentinc.atlassian.net/browse/DEVX-2769

JMX client metrics are broken when running with Docker Compose 2.x, due to problems with the expect container/host name (1.x uses underscores, 2.x uses hyphens).  Solution uses container link aliases to make the hostname predictable for the Prometheus scrape config.

### Author Validation

- [X] ccloud-observability

### Reviewer Tasks

Looking for one reviewer with Docker Compose 1.x and one with Docker Compose 2.x to test works with both version.  Author has 2.x, which highlighted the problem.

- [X] ccloud-observability

### DIFF
--- a/ccloud-observability/docker-compose.yml
+++ b/ccloud-observability/docker-compose.yml
@@ -48,6 +48,9 @@ services:
   prometheus:
     image: prom/prometheus:v2.24.1
     container_name: prometheus
+    links:
+      - producer
+      - consumer
     ports:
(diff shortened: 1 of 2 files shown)


davetroiano (approved):
LGTM, thanks @javabrett

PR #1131, opened 2022-10-08
Patch 1 - Adding absolute path to ccloud.go as GO does not support relative imports.

C-m3-Codin:
### Description 

### Author Validation

### Reviewer Tasks


davetroiano (approved):
LGTM, thank you!

PR #1132, opened 2022-10-12
Make CC Java Client Tutorial Commands Runnable

Robbie-Palmer:
### Description 

On Confluent Cloud, the Java Client tutorial suggests creating the file `~/.confluent/java.config` and gives copyable commands for running an example producer and consumer

Producer:`mvn exec:java -Dexec.mainClass="io.confluent.examples.clients.cloud.ProducerExample" \
-Dexec.args="~/.confluent/java.config test1"`

Consumer: `mvn exec:java -Dexec.mainClass="io.confluent.examples.clients.cloud.ConsumerExample" \
-Dexec.args="~/.confluent/java.config test1"`

This throws an error because `~` is not converted into the path to the user's home directory
`[ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:1.2.1:java (default-cli) on project clients-example: An exception occured while executing the Java class. null: InvocationTargetException: ~/.confluent/java.config not found.`

I have updated the Java Cloud Client examples to support `~` in the config path so the copyable tutorial commands will work

### Author Validation

- [x] clients/cloud

### Reviewer Tasks

- [ ] clients/cloud

### DIFF
--- a/clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/ConsumerAvroExample.java
+++ b/clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/ConsumerAvroExample.java
@@ -15,21 +15,14 @@
  */
 package io.confluent.examples.clients.cloud;
 
-import org.apache.kafka.clients.consumer.ConsumerRecord;
-import org.apache.kafka.clients.consumer.ConsumerRecords;
-import org.apache.kafka.clients.consumer.ConsumerConfig;
-import org.apache.kafka.clients.consumer.Consumer;
-import org.apache.kafka.clients.consumer.KafkaConsumer;
 import io.confluent.kafka.serializers.KafkaAvroDeserializer;
 import io.confluent.kafka.serializers.KafkaAvroDeserializerConfig;
+import org.apache.kafka.clients.consumer.*;
 
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.nio.file.Files;
-import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.Properties;
+
+import static io.confluent.examples.clients.cloud.Util.loadConfig;
 
 public class ConsumerAvroExample {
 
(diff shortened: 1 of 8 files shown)


clients/cloud/java/src/main/java/io/confluent/examples/clients/cloud/ConsumerAvroExample.java
```diff
@@ -15,22 +15,15 @@
...
 package io.confluent.examples.clients.cloud;
 
-import org.apache.kafka.clients.consumer.ConsumerRecord;
-import org.apache.kafka.clients.consumer.ConsumerRecords;
-import org.apache.kafka.clients.consumer.ConsumerConfig;
-import org.apache.kafka.clients.consumer.Consumer;
-import org.apache.kafka.clients.consumer.KafkaConsumer;
 import io.confluent.kafka.serializers.KafkaAvroDeserializer;
 import io.confluent.kafka.serializers.KafkaAvroDeserializerConfig;
+import org.apache.kafka.clients.consumer.*;
```
davetroiano: is the import wildcarding intentional or did your IDE do it automatically? I'll admit I'm in the "never wildcard imports" camp when code will be read outside of IDE and to minimize likelihood of namespace collisions. The latter probably doesn't apply here for a small standalone example but IMO the former does since docs link to the source.

Robbie-Palmer: Thanks, yeh I agree, was automatically done by my IDE, have removed wildcard imports now

davetroiano (approved):
LGTM - thanks for the fix and refactor!

bbejeck (approved):
LGTM - thanks @Robbie-Palmer for the fix!

Robbie-Palmer:
Thanks for the reviews @davetroiano and @bbejeck
I don't have write permissions so cannot merge, if either of you would like to :)

bbejeck:
@Robbie-Palmer could you open the same PR against `7.3.0-post` ?

Robbie-Palmer:
@bbejeck opened now in #1143

PR #1139, opened 2022-11-01
Update pom.xml

ronaldpetty:
Hi. Thanks for the examples. I am not 100% sure, but I suppose the log4j bugs are what caused these changes.

This tutorial:

https://docs.confluent.io/platform/current/schema-registry/schema_registry_onprem_tutorial.html#create-the-transactions-topic

does not seem to work with the existing pom.xml as the log4j packages moved / versioned. I am actually not sure why the confluent-log4j worked only with a version being set (instead of just taking the latest, maybe a Maven pro can confirm).

Thanks again.

### Description 

### Author Validation

### Reviewer Tasks

### DIFF
--- a/clients/avro/pom.xml
+++ b/clients/avro/pom.xml
@@ -70,12 +70,13 @@ http://maven.apache.org/xsd/maven-4.0.0.xsd">
     </dependency>
     <dependency>
        <groupId>org.slf4j</groupId>
-       <artifactId>slf4j-log4j12</artifactId>
+       <artifactId>slf4j-reload4j</artifactId>
    </dependency>
    <!-- Use a repackaged version of log4j with security patches. Default log4j v1.2 is a transitive dependency of slf4j-log4j12, but it is excluded in common/pom.xml -->
     <dependency>
         <groupId>io.confluent</groupId>
         <artifactId>confluent-log4j</artifactId>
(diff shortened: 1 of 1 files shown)


clients/avro/pom.xml
```diff
@@ -70,12 +70,13 @@ http://maven.apache.org/xsd/maven-4.0.0.xsd">
...
     <dependency>
        <groupId>org.slf4j</groupId>
-       <artifactId>slf4j-log4j12</artifactId>
+       <artifactId>slf4j-reload4j</artifactId>
    </dependency>
    <!-- Use a repackaged version of log4j with security patches. Default log4j v1.2 is a transitive dependency of slf4j-log4j12, but it is excluded in common/pom.xml -->
     <dependency>
         <groupId>io.confluent</groupId>
         <artifactId>confluent-log4j</artifactId>
+        <version>1.2.17-cp10</version>
```
davetroiano, line 79: you _should_ be able to remove the `confluent-log4j` dependency and use this instead:

```
    <dependency>
      <groupId>ch.qos.reload4j</groupId>
      <artifactId>reload4j</artifactId>
      <version>${reload4j.version}</version>
    </dependency>
```

davetroiano (changes requested):
thank you for catching this @ronaldpetty!

PR #1171, opened 2022-12-20
ccloud-observability example fixes

davetroiano:
### Description 

[asana](https://app.asana.com/0/1201906632362444/1203565326348401/f)

### Author Validation

manual validation
 - [x] ccloud-observability


ccloud-observability/start.sh
```diff
@@ -63,7 +63,8 @@ fi
 echo "$OUTPUT" | jq .
 export METRICS_API_KEY=$(echo "$OUTPUT" | jq -r ".key")
 export METRICS_API_SECRET=$(echo "$OUTPUT" | jq -r ".secret")
-export CLOUD_CLUSTER=$CLUSTER
+export CLOUD_CLUSTER=$(confluent kafka cluster describe -o json |  jq -c '[.id]')
+export LAG_EXPORTER_ID=$CLUSTER
```
bbejeck, line 67: For my own curiosity - what is the `$CLUSTER`  vs. the `$CLOUD_CLUSTER`? I would think both would be the cluster id

bbejeck (approved):
LGTM

ccloud-observability/start.sh
davetroiano: yeah they basically are the same. It's just a difference in format.  `CLOUD_CLUSTER` and `CLOUD_SCHEMA_REGISTRY` are arrays like `["lkc-123456"]` and `CLUSTER` / `LAG_EXPORTER_ID` is just a string. I tried to have one format and then convert but ran into limitations on environment variable handling, so having a separate ID for the lag exporter was easiest.

PR #1224, opened 2023-06-28
remove obsolete Confluent CLI example

davetroiano:
This example no longer recommends best practice for interacting with Confluent Cloud as it uses `confluent local services kafka consume --could ...` instead of `confluent kafka topic consume ...`

The previously public documentation redirects to this newer getting started guide: https://docs.confluent.io/confluent-cli/current/overview.html

This guide lives in this repository:
https://github.com/confluentinc/docs-confluent-cli, so we can safely delete these examples.

### Description 

[asana](https://app.asana.com/0/1201906632362444/1204738563449396/f)


bbejeck (approved):
LGTM

PR #1225, opened 2023-07-05
🎁 Fixing some issues with p8s and docker compose

bleporini:
### Description 

`start.sh` was failing because of various issues:
- docker compose yaml parser was updated in 2.17 (cf. https://github.com/docker/compose/issues/10407 ) and is not supporting anymore more than one '<<' merge key 
- `confluent connect list` is not a valid CLI command, and as a side effect, it makes the p8s config file generation failing, has a result it has no target at all. 

### Author Validation

I've run multiple `start.sh` --> `stop.sh` cycles on my laptop and on a colleague's one.
 - [ x ] ccloud-observability

### Reviewer Tasks

 - [ x ] ccloud-observability

### DIFF
--- a/ccloud-observability/docker-compose.yml
+++ b/ccloud-observability/docker-compose.yml
@@ -60,19 +60,17 @@ services:
       - ./monitoring_configs/grafana/provisioning/:/etc/grafana/provisioning/
 
   producer:
-    <<: *client-base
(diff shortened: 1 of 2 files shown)


davetroiano (approved):
LGTM thanks @bleporini!

PR #1226, opened 2023-07-11
Removes cleanup command from run.sh after running rbac example

kc596:
### Description 

Removes cleanup after running rbac script `run.sh` as this voids above commands.

### Author Validation

[X] security/rbac

### Reviewer Tasks

### DIFF
--- a/security/rbac/scripts/run.sh
+++ b/security/rbac/scripts/run.sh
@@ -43,5 +43,3 @@ sleep 1
 ./enable-rbac-rest-proxy.sh
 ./enable-rbac-ksqldb-server.sh
 ./enable-rbac-control-center.sh
-
-# ./cleanup.sh
--- a/security/rbac/scripts/run.sh
+++ b/security/rbac/scripts/run.sh
@@ -44,4 +44,4 @@ sleep 1
 ./enable-rbac-ksqldb-server.sh
 ./enable-rbac-control-center.sh
 
-./cleanup.sh
+# ./cleanup.sh


security/rbac/scripts/run.sh
```diff
@@ -44,4 +44,4 @@ sleep 1
 ./enable-rbac-ksqldb-server.sh
 ./enable-rbac-control-center.sh
 
-./cleanup.sh
+# ./cleanup.sh
```
davetroiano: Looking back at history, `run.sh` was always intended to clean up after itself, but I don't see why it couldn't leave services running. That would also work better with the [docs page](https://docs.confluent.io/platform/current/tutorials/examples/security/rbac/docs/index.html#run-example-on-local-install-of-cp) that goes with the example since the step after `Run example` is `Stop example`

just delete the line rather than comment out though

davetroiano (changes requested):

security/rbac/scripts/run.sh
bbejeck: Yep I agree with @davetroiano

bbejeck (commented):
This should go against the `7.4.0-post` branch

security/rbac/scripts/run.sh
kc596: cleanup should be done separately for actually "running" and playing with the examples.

Deleted the line as suggested. Commented initially because it was like that in previous history.

kc596:
> This should go against the `7.4.0-post` branch

Will raise separate PR for it. Because the branch history is different now.

PR #1228, opened 2023-08-01
Clarify use of docker compose vs. docker-compose in Multi-region tutorial

londoncalling:
### Description 

Field feedback.

Docker compose commands are version specific, so if you have an older version of Docker, the newer version of the CLI won't work. 

### Staging server

- http://staging-docs-independent.confluent.io.s3-website-us-west-2.amazonaws.com/docs-platform/PR/2734/7.4/tutorials/examples/multiregion/docs/multiregion.html#prerequisites

### Reviewer Tasks

 [ ] Documentation review

### DIFF
--- a/multiregion/docs/multiregion.rst
+++ b/multiregion/docs/multiregion.rst
@@ -112,6 +112,13 @@ Topic
 -  ``--replica-placement <path-to-replica-placement-policy-json>``: at
    topic creation, this argument defines the replica placement policy for a given
    topic
+   
+Docker commands
+---------------
+
+This version of the tutorial shows ``docker compose`` commands. Earlier versions of the Docker Compose CLI
+used ``docker-compose`` (hyphenated). If ``docker compose`` doesn't work for you, then ``docker-compose`` should work.
+To learn more, see `https://docs.docker.com/compose/reference/ <https://docs.docker.com/compose/reference/>`__.
 
 Download and run the tutorial
 -----------------------------
--- a/multiregion/docs/multiregion.rst
+++ b/multiregion/docs/multiregion.rst
@@ -113,12 +113,20 @@ Topic
    topic creation, this argument defines the replica placement policy for a given
    topic
    
-Docker commands
----------------
+Prerequisites
+-------------
 
-This version of the tutorial shows ``docker compose`` commands. Earlier versions of the Docker Compose CLI
-used ``docker-compose`` (hyphenated). If ``docker compose`` doesn't work for you, then ``docker-compose`` should work.
-To learn more, see `https://docs.docker.com/compose/reference/ <https://docs.docker.com/compose/reference/>`__.
+- Install `Docker Desktop <https://docs.docker.com/desktop/>`__ (version ``4.0.0`` or later) or `Docker Engine <https://docs.docker.com/engine/install/>`__ (version ``19.03.0`` or later) if you don’t already have it.
+
+- Install the `Docker Compose plugin <https://docs.docker.com/compose/install/>`__ if you don’t already have it. This isn’t necessary if you have Docker Desktop, since it includes Docker Compose.
+
+- Start Docker if it’s not already running, either by starting Docker Desktop or, if you manage Docker Engine with ``systemd``, via `systemctl <https://docs.docker.com/config/daemon/systemd/>`__.
+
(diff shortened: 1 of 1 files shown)


multiregion/docs/multiregion.rst
```diff
@@ -112,6 +112,13 @@ Topic
 -  ``--replica-placement <path-to-replica-placement-policy-json>``: at
    topic creation, this argument defines the replica placement policy for a given
    topic
+   
+Docker commands
+---------------
+
+This version of the tutorial shows ``docker compose`` commands. Earlier versions of the Docker Compose CLI
```
davetroiano: "Earlier versions of the Docker Compose CLI" intended to be "Earlier versions of the tutorial"?

davetroiano (commented):
@londoncalling was this a userback? LGTM as is modulo the comment, or we can consider adding a prereqs section with the same content as the prereqs of a kafka tutorial ([example](https://developer.confluent.io/tutorials/creating-first-apache-kafka-producer-application/kafka.html)). I'm not sure which is friendlier for users who don't have the compose plugin: get the plugin and then all copy/paste commands work, or don't get it and have to add hyphens throughout the tutorial. I think the former is friendlier for me but idk about more generally - you make the call

londoncalling:
@davetroiano Yes, this was a Userback feedback, but from a Confluent employee, here are some thoughts:

- I like the idea of a prerequisites section .. **can I just use all the same prereqs listed in the [Kafka tutorial](https://developer.confluent.io/tutorials/creating-first-apache-kafka-producer-application/kafka.html#prerequisites) or are there some that do not apply?**

- Regarding your comment re: "earlier versions", Docker had actually updated `docker-compose` to `docker compose` when going from v1 to v2 of that product, as noted in their [Docker Compose reference](https://docs.docker.com/compose/reference/), which is why I thought you had updated the tutorial to use the newer version

All that said, I still think it's worth explicitly mentioning and/or referring them to the fact that you can use `docker-compose` or `docker compose` depending on what version of the Docker CLI you have, given Alvin's experience ([UserBack here](https://app.userback.io/viewer/35466/68574/2853290jkF114yy/)):

_Hi Victoria, I've tried using our lab internal training and I can't run the command if I'm using the docker compose. It will run if I use docker-compose. On the other hand, noted, thanks for your prompt response._

davetroiano:
> can I just use all the same prereqs listed in the [Kafka tutorial](https://developer.confluent.io/tutorials/creating-first-apache-kafka-producer-application/kafka.html#prerequisites) or are there some that do not apply?

You can use all the same prereqs.

In looking at that tutorial the other aspect that I like (aside from the docker compose instructions) is that it makes sure docker is actually running - as it stands we don't say this anywhere but we should since many devs don't have Docker configured to start automatically on reboot.

>Regarding your comment re: "earlier versions", Docker had actually updated docker-compose to docker compose when going from v1 to v2 of that product, as noted in their [Docker Compose reference](https://docs.docker.com/compose/reference/), which is why I thought you had updated the tutorial to use the newer version

My confusion on the "Earlier versions of the Docker Compose CLI used ``docker-compose`` (hyphenated)" line is that to me the Docker Compose CLI _is_ `docker-compose`. e.g., it wouldn't be accurate to expand this and say:

>Earlier versions of the Docker Compose CLI used ``docker-compose`` (hyphenated) while the latest version of the Docker Compose CLI uses "docker compose".

How about:

>Earlier versions of Docker did not include Docker Compose and a separate (hyphenated) `docker-compose` CLI was required to use Docker Compose. If your local Docker installation does not support Docker Compose (i.e., `docker compose version` throws an error), then you can either upgrade Docker or install the Compose plugin as documented in the Prerequisites section, or, if you have the standalone `docker-compose` CLI installed and you don't wish to modify your local Docker installation, then you can substitute `docker-compose` wherever `docker compose` is used in the tutorial.

londoncalling:
Hi @davetroiano , please take a look at the staging link in the PR "Description" above and see what you think. If you think the "tip" is too heavy handed, I can just add this info at the end of the Prerequisites without the "tip" formatting. Otherwise, whenever you think this is ready for prime time, can you give me an approval to merge?

davetroiano (approved):
LGTM!

Issue #1262, opened 2024-09-30
Cloud ETL example does not work with Confluent CLI 4.x


metadaddy:
**Description**
I recently tried to work through the [Cloud ETL example](https://github.com/confluentinc/examples/tree/7.5.0-post/cloud-etl), but hit numerous issues due to the [breaking changes in Confluent CLI 4.x](https://docs.confluent.io/confluent-cli/current/migrate.html#cli-breaking-changes-v4). For example, `ccloud::enable_schema_registry()` invokes a command of the form:

```
confluent schema-registry cluster enable --cloud aws --geo us
```

which results in an error:

```
Error: unknown flag: --cloud
Usage:
  confluent schema-registry cluster [command]

Available Commands:
  describe    Describe the Schema Registry cluster for this environment.
  update      Update global mode or compatibility of Schema Registry.

Global Flags:
  -h, --help            Show help for this command.
      --unsafe-trace    Equivalent to -vvvv, but also log HTTP requests and responses which might contain plaintext secrets.
  -v, --verbose count   Increase verbosity (-v for warn, -vv for info, -vvv for debug, -vvvv for trace).

Use "confluent schema-registry cluster [command] --help" for more information about a command.
```

The [documentation states that CLI 3.0.0 or later is required](https://docs.confluent.io/cloud/current/get-started/tutorials/cloud-etl.html#local-tools), but this is no longer accurate. Also, at the time, I was unable to find a location from which I could download a 3.x.x version of the CLI.

I [figured out the changes to get it working with CLI 4.4.0](https://github.com/backblaze-b2-samples/confluent-examples/commit/918cc6d5dc3ce4408c72ed4dc62d49124556370c), but my scope was limited to the Cloud ETL example, so I'm hesitant to submit a PR. Unfortunately, I don't have the time or knowledge to migrate all of the examples to CLI 4.x.

I'm guessing (hoping?) that someone at Confluent is on the case, and already migrating the examples to CLI 4.x, so this is mostly to help anyone else that encounters the same issues, and point the way to the [Confluent CLI v3.65.0 download links on GitHub](https://github.com/confluentinc/cli/releases/tag/v3.65.0) that I since found (assuming that CLI version works!).

In the interim, you might also want to update the docs to state that _only_ 3.x.x CLI versions work, rather than 3.0.0 and above, and include the link to download Confluent CLI v3.65.0.

[...]

PR #1264, opened 2024-10-08
Allow S3-Compatible Cloud Object Storage in CloudETL Example

metadaddy:
### Description 

This change allows the use of an S3-compatible cloud object store such as [Backblaze B2](https://www.backblaze.com/cloud-storage) with the CloudETL example as an alternative to Amazon S3.

With this change, the user may configure an AWS profile with `endpoint_url` set to a value such as `https://s3.us-west-004.backblazeb2.com` in the `config` file. For example:

```
[profile b2-example]
endpoint_url = https://s3.us-west-004.backblazeb2.com
region = us-west-004
```

The `setup_s3_storage.sh` script reads this value (via `aws configure get endpoint_url --profile $S3_PROFILE`) into a new `S3_ENDPOINT_URL` environment variable which is used as `store.url` for the connectors. If the endpoint URL is not set in the profile, then the Amazon S3 global default, `https://s3.amazonaws.com`, is used.

There is also a fix to `read-data.sh` - the `list-objects` call was missing the `--profile $S3_PROFILE`, so it incorrectly used the default profile.

### Author Validation

[x] cloud-etl

### Reviewer Tasks

[ ] cloud-etl

### DIFF
--- a/cloud-etl/connectors/s3_avro.json
+++ b/cloud-etl/connectors/s3_avro.json
@@ -6,6 +6,7 @@
     "kafka.api.secret": "$CLOUD_SECRET",
     "aws.access.key.id": "$AWS_ACCESS_KEY_ID",
     "aws.secret.access.key": "$AWS_SECRET_ACCESS_KEY",
+    "store.url": "$S3_ENDPOINT_URL",
     "s3.bucket.name": "$S3_BUCKET",
     "data.format": "AVRO",
     "time.interval" : "HOURLY",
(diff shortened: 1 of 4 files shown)


metadaddy:
I signed the CLA, but for some reason, it doesn't seem to show up here.

metadaddy:
Hi @davetroiano - you replied to me on #1262 last month, pointing to the PR for the CLI 4 compatibles samples. Any interest in taking this on? It shows how you can send data to _any_ S3-compatible cloud object store (Backblaze, Minio, IBM Cloud Object Storage, etc etc) by setting the endpoint and region in the AWS profile (which you've likely already done if you're using any of the AWS SDKs/CLI with one of those providers).

PR #1296, opened 2026-08-10
Fix integer overflow in vendored json-parser allocation size

Socialpranker:
This patch was written by Claude Sonnet 5 (Anthropic); I reviewed the change and the reasoning before submitting.

clients/cloud/c/json.c vendors the json-parser library. In new_value(), the array and object branches multiply value->u.array.length and value->u.object.length by sizeof(json_value*) and sizeof(json_object_entry) respectively before passing the result to json_alloc(). Neither multiplication is checked for overflow first. If it wraps, json_alloc() receives a small size, the allocation succeeds, and the parser then writes length elements into a buffer that is too small for them, a heap buffer overflow driven entirely by the JSON input being parsed.

json_alloc() does guard state->used_memory against wrapping via state->ulong_max, but that check only runs after the multiplication has already happened, so it cannot catch an overflow in the multiplication itself.

The fix adds a bound check in both branches before the multiplication, using the same state->ulong_max field json_alloc() already computes for the same purpose, so the guard follows the existing convention in the file rather than introducing a new one. On failure it returns 0, which is exactly how allocation failure is already signaled everywhere else in new_value(). The diff is intentionally limited to these two checks, no other code in the file was touched.

I compiled clients/cloud/c/json.c standalone with gcc -Wall -Wextra -std=c99 before and after the change; the warning output is identical (two pre-existing unused-parameter warnings in default_alloc/default_free, unrelated to this fix). I did not build consumer.c/producer.c through the Makefile because that requires librdkafka, which isn't installed in the environment I used, but json.c has no dependency on librdkafka and compiles cleanly on its own.

### DIFF
--- a/clients/cloud/c/json.c
+++ b/clients/cloud/c/json.c
@@ -41,6 +41,7 @@ const struct _json_value json_value_none;
 #include <string.h>
 #include <ctype.h>
 #include <math.h>
+#include <limits.h>
 
 typedef unsigned int json_uchar;
 
@@ -138,7 +139,7 @@ static int new_value (json_state * state,
             if (value->u.object.length == 0)
                break;
 
-            if (value->u.object.length > state->ulong_max / sizeof (*value->u.object.values))
+            if (value->u.object.length > (int) (INT_MAX / sizeof (*value->u.object.values)))
                return 0;
 
             values_size = sizeof (*value->u.object.values) * value->u.object.length;
--- a/clients/cloud/c/json.c
+++ b/clients/cloud/c/json.c
@@ -121,6 +121,9 @@ static int new_value (json_state * state,
             if (value->u.array.length == 0)
                break;
 
+            if (value->u.array.length > state->ulong_max / sizeof (json_value *))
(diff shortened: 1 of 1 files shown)


Socialpranker:
Pushed a follow-up commit: the object-branch guard checked against `ulong_max`, but `values_size` that stores the multiplication result is declared as a plain `int`. A length between `INT_MAX/sizeof(entry)` and `ulong_max/sizeof(entry)` would pass that guard and still overflow `values_size` itself. Fixed to guard against `INT_MAX/sizeof(entry)` instead, matching the type that actually holds the product. The array branch was already correct since it allocates directly with no intermediate int.

Socialpranker:
Closing this as stale on my side — no reflection on the change itself.

The default branch is currently driven entirely by automation, and the last human-merged pull request was in August 2025. I'd rather not leave a PR sitting in a queue that isn't being processed.

The fix itself remains valid and the branch stays up. If maintenance picks up and this is still useful, feel free to reopen — no need to ask me first.

